PR #16, opened 2025-08-18
fix: don’t run test workflow on docs

Poggen:


kejne:
This should be solved with #19 already, but thanks for you contribution!

PR #30, opened 2025-09-05
feat: add jetstream cross-account examples

kejne:
A typical scenario for NAuth would be that two systems integrate with each other and that they are exporting/importing to one another.

JetStreams are a bit complicated to export/import since it requires the exports of specific JS API:s.

This adds an example of a cross-account solution to the development environment.


jotu (approved):
LGTM

Issue #26, opened 2025-08-26
Limit access to reading and updating kubernetes secrets to allowed namespaces and secrets


kejne:
Currently, accounts and users will also generate kubernetes secrets for the seed of the nkeys and the user creds.

Since Nauth has access to update the secrets over the entire cluster, additional guard rails need to be in place.

We want to limit Nauth to only interact with secrets that are "owned" by Nauth itself. We need to mark the nauth secrets with some information (label) to that these secrets were generated/ managed by the Nauth operator.

Depends on: #46  

# Proposed actions
- [x] Any interaction with secrets in nauth must require secrets and holding namespaces to be labeled with a specific label that allows updates
- [x] Nauth sets labels on secrets that it creates

PR #48, opened 2025-10-29
feat(secrets)!: add and rely on a "managed" label for spawned secrets

cmoscofian:
Nauth stores accounts private key and signing key seeds and user credentials as
secrets within the system's namespace. In order to guardrail nauth from mutating
non-nauth secrets secrets it now relies on the *nauth.io/managed* label.

Also, relies on the labels for fetching external required secrets (operator user
credentials and signing key), instead of relying on named patterns.

Signed-off-by: Charles Moscofian <charlesmoscofian@hotmail.com>

---
Issue: #26

#### Operator secrets labels
2025-10-29T16-23-59
2025-10-29T16-23-04

#### Nauth secrets labels
2025-10-29T16-22-25
2025-10-29T16-21-43

By submitting a request, you represent that you have the right to license
licensed under the [License](LICENSE).

### DIFF
--- a/internal/adapter/out/k8s/secret.go
+++ b/internal/adapter/out/k8s/secret.go
@@ -61,7 +61,7 @@ func (k SecretStorer) ApplySecret(ctx context.Context, owner *ports.SecretOwner,
 			return fmt.Errorf("failed to create secret: %w", err)
 		}
 	} else {
-		// TODO: require nauth.io/managed: "true" label in release v0.1.3
+		// TODO: require nauth.io/managed: "true" label in future release after explicit addition of new labels
 		if currentSecret.Labels == nil {
 			currentSecret.Labels = make(map[string]string)
 		}


internal/core/domain/constants.go
```diff
@@ -4,22 +4,26 @@ const (
...
 )
 
 const (
-	SecretTypeAccountRoot = "ac-root"
-	SecretTypeAccountSign = "ac-sign"
+	SecretTypeAccountRoot     = "ac-root"
+	SecretTypeAccountSign     = "ac-sign"
+	SecretTypeOperatorSign    = "op-sign"
+	SecretTypeOperatorCreds   = "op-creds"
+	SecretTypeUserCredentials = "us-creds"
```
kejne: Can we take the chance now and abbreviate a little less?
"user", "account", "operator" are not that long and make it much easier to read.

internal/adapter/out/k8s/secret.go
```diff
@@ -38,38 +38,39 @@ func NewK8sSecretStorer(client client.Client) *SecretStorer {
...
+
 		if err := k.client.Create(ctx, newSecret); err != nil {
 			return fmt.Errorf("failed to create secret: %w", err)
 		}
 	} else {
-		newSecret = currentSecret
-		newSecret.StringData = valueMap
+		if currentSecret.Labels == nil {
+			currentSecret.Labels = make(map[string]string)
+		}
```
kejne: I assume that we're simply overwriting the secret without labels now in order to be able to migrate?
Would be good with a TODO here, since this is where I would expect a check that the secret is indeed managed by nauth before updating it.

cmd/main.go
```diff
@@ -206,11 +206,11 @@ func main() {
 		os.Exit(1)
 	}
 
-	natsUrl := os.Getenv("NATS_URL")
+	natsURL := os.Getenv("NATS_URL")
```
thobiaskarlsson: **praise**: Nice to clean up faulty abbreviation casing while you're at it. ⭐

internal/core/service/account.go
```diff
@@ -97,8 +97,9 @@ func (a *AccountManager) CreateAccount(ctx context.Context, state *natsv1alpha1.
 		Name:      getAccountRootSecretName(state.GetName(), accountPublicKey),
 		Namespace: state.GetNamespace(),
 		Labels: map[string]string{
-			domain.LabelAccountId:         accountPublicKey,
-			domain.LabelAccountSecretType: domain.SecretTypeAccountRoot,
+			domain.LabelAccountId:  accountPublicKey,
+			domain.LabelSecretType: domain.SecretTypeAccountRoot,
+			domain.LabelManaged:    "true",
```
thobiaskarlsson: **question**: Would it make sense to extract this fixed `nauth.io/managed` label value to a constant as well? Suggestion `domain.LabelManagedValue`.

thobiaskarlsson (approved):
LGTM.

examples/nauth/manifests/operator.yaml
```diff
@@ -12,6 +12,8 @@ kind: Secret
 type: Opaque
 metadata:
   name: operator-op-sign
+  labels:
+    nauth.io/secret-type: op-sign
```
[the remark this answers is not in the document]
kejne: Yes, some documentation on what `managed: true` means would be good.

internal/core/domain/constants.go
kejne: There was actually a reason behind shortening the names of the secret since the maximum length can only be 64 chars long and we append after the name of whatever the account/user is named.

suggestion: do the refactoring of secret name *after* the updates have been rolled out when all secrets are looked up by label instead.

examples/nauth/manifests/operator.yaml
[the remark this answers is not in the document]
cmoscofian: I'm not against it it, but I don't really see the benefit of documenting `nauth.io/managed: true` since it's for internal usage only. I was trying to see how other projects take the approach and I did not see this being documented, [this is the only mention I found from external secrets for example](https://github.com/external-secrets/external-secrets/blob/612d5dbe601e5f5c8a89c92db12ab7e8ff714753/pkg/controllers/common/common.go#L39).
It might be good to document and be on the safe side, but I don't see anyone just randomly labeling a secret with the nauth.io domain.

The only real usecase I can imagine happening is someone bringing resources back  from a backup.

internal/core/domain/constants.go
cmoscofian: I agree, this particular abbreviation rubbed me the wrong way. I'll rename it all to the following:
```go
	SecretTypeAccountRoot              = "account-root"
	SecretTypeAccountSign              = "account-sign"
	SecretTypeOperatorAdminSign        = "admin-sign"
	SecretTypeOperatorAdminCredentials = "admin-creds"
	SecretTypeUserCredentials          = "user-creds"
```

examples/nauth/manifests/operator.yaml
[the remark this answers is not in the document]
thobiaskarlsson: Documentation will be done in a later stage.

internal/core/domain/constants.go
thobiaskarlsson: Done.

Issue #62, opened 2025-11-21
Add support for namespaced deployment


wingsofovnia:
Currently, it is not possible to deploy NAuth scoped to a single namespace, like [we can](https://github.com/nats-io/k8s/blob/223646f12ada1fd36fc88da33408e45f8d438552/helm/charts/nack/values.yaml#L55-L56) NACK.

The problem is two-fold.

## [`crds.install`] doesn't do what it says

From [`values.yaml`](https://github.com/WirelessCar/nauth/blob/7283990c0ab2037151cb7161b28795767e4a8b89/charts/nauth/values.yaml#L109-L110):

> If [`crds.install`] set to `false`, the CRDs will only be installed **if they do not already exist**. 

This is however not happening because when set to `false`, it will not be installed regardless, as `templates/crds.yaml` skips all crds completely if `false`:
https://github.com/WirelessCar/nauth/blob/7283990c0ab2037151cb7161b28795767e4a8b89/charts/nauth/templates/crds.yaml#L1-L11

```
❯ helm install nauth oci://ghcr.io/wirelesscar/nauth \
    --set crds.install=false \
    --create-namespace \
    --namespace nauth
Pulled: ghcr.io/wirelesscar/nauth:0.2.0
Digest: sha256:2d2b8c808c666c3cb775a1cd21dc569901d19c15c971ffed6235fd00bf6d908b
NAME: nauth
LAST DEPLOYED: Fri Nov 21 16:48:54 2025
NAMESPACE: nauth
STATUS: deployed
REVISION: 1
TEST SUITE: None
❯ k get crds | grep nauth
[empty]
```

## RBAC is cluster-scoped
Even if this is fixed, it is still not possible to install a copy of nauth to another namespace due to the use of a `ClusterRole`:
```
❯ helm install nauth oci://ghcr.io/wirelesscar/nauth --set crds.install=false --create-namespace --namespace nauth-2
Pulled: ghcr.io/wirelesscar/nauth:0.2.0
Digest: sha256:2d2b8c808c666c3cb775a1cd21dc569901d19c15c971ffed6235fd00bf6d908b
Error: INSTALLATION FAILED: Unable to continue with install: ClusterRole "account-admin-role" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "nauth-2": current value is "nauth"
```

# Proposal
[...]

PR #64, opened 2025-11-24
feat(#62): Add support for namespaced deployments

wingsofovnia:
* Add `--namepsace` arg to configure controller to watch and manage resources in a single namespace ([ref](https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#watching-resources-in-specific-namespaces)).
* Add `.Values.namespaced` to enable only watching resources in the namespace it is deployed, [similar to nack](https://github.com/nats-io/k8s/blob/223646f12ada1fd36fc88da33408e45f8d438552/helm/charts/nack/values.yaml#L55-L56).
* ~Fix `crds.install` to do what it says in case it is set to `false` ~ install CRDS if they do not already exist. Cover it with helm-unittest.~
* Do not use Cluster-wide RBAC for namespaced deployment when `.Values.namespaced` ~ true

Closes https://github.com/WirelessCar/nauth/issues/62

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@
 name: Operator tests
 
 on:
-  pull_request:
+  pull_request_target:
     paths:
       - 'dir/**'
       - 'cmd/**'
@@ -37,6 +37,8 @@ jobs:
     steps:
       - name: Clone the code
         uses: actions/checkout@v4
+        with:
+          ref: "${{ github.head_ref || github.ref }}"
 
       - name: Setup Go
         uses: actions/setup-go@v5


charts/nauth/templates/crds.yaml
```diff
@@ -1,11 +1,15 @@
...
+  {{- $manifest := $.Files.Get $path | fromYaml }}
+  {{- if not $.Values.crds.install }}
+    {{- if lookup $manifest.apiVersion $manifest.kind "" $manifest.metadata.name }}
+      {{- continue }}
+    {{- end }}
+  {{- end }}
+  {{- if $.Values.crds.keep }}
+    {{- $newAnnotations := dict "helm.sh/resource-policy" "keep" | merge $manifest.metadata.annotations }}
+    {{- $_ := set $manifest.metadata "annotations" $newAnnotations }}
+  {{- end }}
```
kejne: **issue:** Introducing Helm features like `lookup` will break the idea of being able to utilize [Rendered Manifest Pattern](https://akuity.io/blog/the-rendered-manifests-pattern). The only problem currently is that the documentation is written in a way which suggests that the code has logic, whereas I believe the intention of the writer was that the crd:s would have already been provided.

.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@
 name: Operator tests
 
 on:
-  pull_request_target:
+  pull_request:
```
kejne: **praise:** Great, this would make much more sense for a test workflow like this.

charts/nauth/templates/crds.yaml
```diff
@@ -1,11 +1,15 @@
...
-{{- $manifest := $.Files.Get $path | fromYaml }}
-{{- if $.Values.crds.keep }}
-{{- $newAnnotations := dict "helm.sh/resource-policy" "keep" | merge $manifest.metadata.annotations }}
-{{- $_ := set $manifest.metadata "annotations" $newAnnotations }}
-{{- end }}
-{{- $manifest | toYaml  }}
-{{- print "\n---\n" }}
-{{- end }}
+  {{- $manifest := $.Files.Get $path | fromYaml }}
+  {{- if not $.Values.crds.install }}
```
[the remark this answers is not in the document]
kejne: Let's skip updates to the behavior of how crds are installed since this should already cover the use case with a declarative approach.

kejne:
> Thank you for looking into it. Reverted the crds changes and only updated the doc string to be less confusing (also aligned the comments to be recognisable by `helm-docs`). Ready for another pass.

Great! Could you just remove the github actions update in this PR as well?
Will be updating the workflows a bit more as a whole since this will only solve a part.

kejne (approved):
LGTM!

PR #69, opened 2025-11-27
chore(chart): add nauth-crds complementary crds-only chart

wingsofovnia:
In https://github.com/WirelessCar/nauth/issues/62, we discussed a use case when it is currently not possible to use `nauth` as a dependency chart when also using those CRDs in parent:

```
❯ helm create my-chart
Creating my-chart
❯ cd my-chart

❯ echo "dependencies:
  - name: nauth
    version: \"0.2.0\"
    repository: \"oci://ghcr.io/wirelesscar\"" >> Chart.yaml
❯ helm dependency update

❯ curl -L \
  https://raw.githubusercontent.com/WirelessCar/nauth/refs/heads/main/examples/nauth/manifests/scenarios/simple-account/simple-account.yaml \
  -o templates/simple-account.yaml

❯ minikube start
❯ helm install my-chart-release .
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "example-account" namespace: "simple-account" from "": no matches for kind "Account" in version "nauth.io/v1alpha1"
ensure CRDs are installed first, resource mapping not found for name: "example-user" namespace: "simple-account" from "": no matches for kind "User" in version "nauth.io/v1alpha1"
ensure CRDs are installed first]
```

This happens because CRDs rendered via `templates/` are installed at the same time as everything else. Kubernetes doesn’t register them in time for CR validation unlike it would with `crds/` folder.

In that issue a suggestion was to create a separate chart with crds only.

This change adds a complementary `nauth-crds` that contains only crds and plays along well with the main chart with `crds.install=false`.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -96,12 +96,15 @@ jobs:
         run: |
           VERSION=$(echo ${{ github.ref_name }} | sed 's/^v//')
           echo "version=${VERSION#v}" >> $GITHUB_OUTPUT
-      - name: Helm package
+      - name: Helm package nauth
         run: helm package nauth
-      - name: login to gcr using helm
+      - name: Helm package nauth-crds
+        run: helm package nauth-crds
+      - name: Authenticate Helm to ghcr.io
         run: |
-          echo ${{ secrets.GITHUB_TOKEN }} | helm registry login ghcr.io/wirelesscar/nauth --username ${{ github.repository_owner }} --password-stdin
-      - name: helm push
-        run: |
-          helm push nauth-${{ steps.semver.outputs.version }}.tgz oci://ghcr.io/wirelesscar
+          echo ${{ secrets.GITHUB_TOKEN }} | helm registry login ghcr.io --username ${{ github.repository_owner }} --password-stdin
+      - name: Helm push nauth
+        run: helm push nauth-${{ steps.semver.outputs.version }}.tgz oci://ghcr.io/wirelesscar
+      - name: Helm push nauth-crds
+        run: helm push nauth-crds-${{ steps.semver.outputs.version }}.tgz oci://ghcr.io/wirelesscar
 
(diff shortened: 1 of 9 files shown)


.github/workflows/release.yaml
```diff
@@ -96,12 +96,15 @@ jobs:
...
         run: |
-          echo ${{ secrets.GITHUB_TOKEN }} | helm registry login ghcr.io/wirelesscar/nauth --username ${{ github.repository_owner }} --password-stdin
-      - name: helm push
-        run: |
-          helm push nauth-${{ steps.semver.outputs.version }}.tgz oci://ghcr.io/wirelesscar
+          echo ${{ secrets.GITHUB_TOKEN }} | helm registry login ghcr.io --username ${{ github.repository_owner }} --password-stdin
+      - name: Helm push nauth
+        run: helm push nauth-${{ steps.semver.outputs.version }}.tgz oci://ghcr.io/wirelesscar
+      - name: Helm push nauth-crds
+        run: helm push nauth-crds-${{ steps.semver.outputs.version }}.tgz oci://ghcr.io/wirelesscar
```
kejne: - **issue:** The registry is immutable, which requires both charts to be updated or the release will fail. Since the chart version is set in the Chart.yaml this is likely to happen. Great however that the crds (which is the chart most likely to fail) is last, so the workflow will be marked as failed but actual chart released. I'm OK with this issue ATM since I think we can come up with a better solution later on.

- **praise:** Really like that you took the effort in updating all the locations needed for releasing a new chart. 🏅

kejne (approved):
Great that you've updated all the locations which would be impacted by the updates!

Found some issues this will bring with it, but I believe that these can be refined separately.

.github/workflows/release.yaml
kejne: Yes, the workflows are quite rudimentary and more or less thrown together.
Also the reason that this PR was merged - it does the job and would be good to update the functionality.
I believe that the `0.0.0-snapshot` approach and using the tag as the master a viable option going forward.

PR #82, opened 2025-12-03
feat: support importing existing accounts in observer-only mode

### LINKED ISSUE #67: Cannot create users for the system account
### Problem

NAuth requires the system account user credentials and operator signing key to be provided. This means the system account must already exist before NAuth can function and not managed by NAuth

However, to create a `User` CRD, you must reference an `Account` CRD via `spec.accountName`.

**The gap**: Since the system account is not managed by NAuth, there is no corresponding `Account` CRD. This creates an impossible situation where:
   - The system account exists (required for NAuth to function)
   - But you cannot create users for it (because no Account CRD exists)

This limitation prevents creating users in the system account, which is especially problematic for NATS monitoring tools that need system-level access (e.g., subscribing to `$SYS.>` subjects).

##### Option 1. Auto-create Account CRD on bootstrap
When NAuth starts up and detects the system account credentials secret, automatically create an `Account` CRD that represents the existing system account. This would make the system account "visible" to NAuth without requiring manual setup. Can be behind a flag.

##### Option 2. Support account reference by public key
Extend `User` CRD to allow referencing accounts not managed by NAuth.
```
spec:
  accountPublicKey: "ADIGHBZB7RFFIIRZV5UEK2JPAF...."
  accountSigningKeySecretRef:    # core/v1.SecretKeySelector, or custom without "optional", and/or "namespace"
    name: "nats-my-account-sign"
    key: "default"
```

##### Option 3. Support external Account CRDs
Add support for "external" accounts that reference existing accounts.
```
spec:
  externalAccount:
    publicKey: "ADIGHBZB7RFFIIRZV5UE...."
    signingKeySecretRef:    # core/v1.SecretKeySelector, or custom without "optional", and/or "namespace"
      name: "nats-my-account-sign"
      key: "default"
```

In my opinion, `Option 1` is the cleanest solution for the system account use case in terms of effort/simplicity. However, `Option 2` or `Option 3` (alone or combined with `Option 1`) would provide more flexibility for other external accounts beyond just the system account.

### Workaround

I haven't tested it yet, but it might be possible to create a "fake" `Account` CRD that has pre-populated labels and status of a system account. This is fragile, error-prone, and risks NAuth overwriting the account JWT if reconciliation is triggered.

wingsofovnia:
Adds support for importing existing NATS accounts that are not managed by NAuth in a read-only mode so that NAuth can manage their users.

In particular, this change enables user management for the system account* by NAuth.

This change is inspired by the approach Crossplane uses when [importing existing resources](https://docs.crossplane.io/v2.1/guides/import-existing-resources/), which is the behaviour NAuth is [going](https://github.com/WirelessCar/nauth/issues/67#issuecomment-3587358100) towards.

* Account observe mode: Status of `Account` CRs with `nauth.io/management-policy: observe` are imported from NATS rather than created/managed.
* Account JWT gets fetched from NATS via `$SYS.REQ.ACCOUNT.<id>.CLAIMS.LOOKUP` and popularity `Account` CR status.
* System account protection*: prevents updating system accounts to avoid changes in the very sys creds NAuth needs to core functionality (must use observe mode).
* Accounts in observe mode are not deleted from NATS when the CR is removed.
* Mentions the new function in README.md

Closes #67 

### DIFF
--- a/internal/core/service/account_test.go
+++ b/internal/core/service/account_test.go
@@ -246,10 +246,10 @@ var _ = Describe("Account manager", func() {
 			Expect(out.Exports).To(HaveLen(2))
 			var streamExp, svcExp *v1alpha1.Export
 			for _, e := range out.Exports {
-				switch {
-				case e.Name == "stream-exp":
+				switch e.Name {
+				case "stream-exp":
 					streamExp = e
-				case e.Name == "svc-exp":
+				case "svc-exp":
 					svcExp = e
 				}
 			}
@@ -275,10 +275,10 @@ var _ = Describe("Account manager", func() {
 			Expect(out.Imports).To(HaveLen(2))
 			var impStream, impSvc *v1alpha1.Import
 			for _, im := range out.Imports {
-				switch {
-				case im.Name == "imp-stream":
+				switch im.Name {
+				case "imp-stream":
 					impStream = im
-				case im.Name == "imp-svc":
+				case "imp-svc":
 					impSvc = im
 				}
 			}


internal/core/ports/in.go
```diff
@@ -9,6 +9,7 @@ import (
 type AccountManager interface {
 	CreateAccount(ctx context.Context, state *natsv1alpha1.Account) error
 	UpdateAccount(ctx context.Context, state *natsv1alpha1.Account) error
```
kejne: **praise:** thanks for cleaning up! ⭐

kejne (commented):
Thanks for another good PR!
There is some ongoing work with refactoring, so will try to get this in before it comes to this.

The approach is sound and would work as an initial approach to handling life cycles (and may be enough for longer period as well).
In the long run the label will likely be moved into the CRD so it's part of the API, but I believe that might be a while.

I had some performance & security concerns that I commented on that would be nice to look at.

What kind of tests did you do?

internal/adapter/out/nats/nats.go
```diff
@@ -167,3 +202,5 @@ func (n natsClient) getOperatorAdminCredentials(ctx context.Context, namespace s
 	}
 	return creds, nil
 }
+
+var _ ports.NATSClient = (*natsClient)(nil)
```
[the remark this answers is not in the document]
wingsofovnia: sorry, didn't quite get that.
this is to assure `natsClient` implements `ports.NATSClient` at compile time, but I couldn't find this pattern to be used elsewhere in the repo so I'd rather remove it.

internal/core/service/account.go
```diff
@@ -243,8 +248,80 @@ func (a *AccountManager) UpdateAccount(ctx context.Context, state *natsv1alpha1.
...
+	}
+	accountSigningSeed, ok := accountSigningSecret[domain.DefaultSecretKeyName]
+	if !ok {
+		return fmt.Errorf("account sign secret for account %s is malformed", accountID)
+	}
+	accountSigningKeyPair, err := nkeys.FromSeed([]byte(accountSigningSeed))
+	if err != nil {
+		return fmt.Errorf("failed to get account key pair for signing from seed for account %s: %w", accountID, err)
+	}
+	accountSigningPublicKey, _ := accountSigningKeyPair.PublicKey()
```
kejne: **note:** We will likely want to package the loading of needed secrets & validation in a more streamlined way - but it will work nicely for this PR 👌

kejne:
@wingsofovnia just haven't tweaked all the settings for the repo yet, keeping it to most restrictive to start with.
Will release as well.

PR #86, opened 2025-12-11
refactor: flatten package structure

henriropp:
The package structure was unnecessary complicated. This PR simplifies by flattening the package structure. The method names for some structs have been updated where necessary as part of the move to a package.

**New package structure:**
```
internal/
    account/         - Account related logic
    controller/      - Controllers (reconcilers)
    k8s/             - K8S clients
    nats/            - Nats client
    user/            - User related logic
```

By submitting a request, you represent that you have the right to license
licensed under the [License](LICENSE).

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -226,7 +226,7 @@ func main() {
 
 	secretClient := secret.NewClient(mgr.GetClient())
 	accountClient := k8s.NewAccountClient(mgr.GetClient())
-	natsClient := natsc.NewNATSClient(natsURL, secretClient)
+	natsClient := natsc.NewClient(natsURL, secretClient)
 
 	accountManager := account.NewManager(
 		accountClient,
--- a/internal/account/mocks_test.go
+++ b/internal/account/mocks_test.go
@@ -34,7 +34,7 @@ func (s *SecretStorerMock) Get(ctx context.Context, namespace string, name strin
 	return args.Get(0).(map[string]string), args.Error(1)
 }
 
-// GetSecretsByLabels implements ports.SecretStorer.
+// GetByLabels implements ports.SecretStorer.
 func (s *SecretStorerMock) GetByLabels(ctx context.Context, namespace string, labels map[string]string) (*corev1.SecretList, error) {
 	args := s.Called(ctx, namespace, labels)
 	return args.Get(0).(*corev1.SecretList), args.Error(1)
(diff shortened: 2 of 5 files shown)


internal/account/account.go
```diff
@@ -12,22 +12,44 @@ import (
...
-	natsClient     ports.NATSClient
-	secretStorer   ports.SecretStorer
+	accounts       AccountGetter
+	natsClient     NatsClient
+	secretStorer   SecretStorer
 	nauthNamespace string
 }
 
-func NewAccountManager(accounts ports.AccountGetter, natsClient ports.NATSClient, secretStorer ports.SecretStorer, opts ...func(*AccountManager)) *AccountManager {
+func NewAccountManager(accounts AccountGetter, natsClient NatsClient, secretStorer SecretStorer, opts ...func(*AccountManager)) *AccountManager {
```
kejne: **suggestion:** if using a package like `account` it makes sense to remove the extra `Account` in exported funcs and structs. For example:
```go
// Now
account.NewAccountManager(...)
account.AccountManager{}

// Proposal
account.NewManager(...)
account.Manager{}
```

internal/k8s/secret.go
```diff
@@ -19,28 +16,32 @@ import (
...
+
+type SecretClient struct {
 	client              client.Client
 	controllerNamespace string
 }
 
-func NewK8sSecretStorer(client client.Client) *SecretStorer {
-	k8sSecretStorer := &SecretStorer{}
+func NewSecretClient(client client.Client) *SecretClient {
+	secretClient := &SecretClient{}
```
kejne: **suggestion:** Maybe it would make more sense to move this into package `secret` instead? The `k8s` package came from the hexagonal approach where `k8s` was *one of the targets*. This would put the secret labels align better (since some labels reside in the controller)

kejne (changes requested):
Overall nice!

Added some optional suggestions.
Make sure the PR has relevant information in it with intent and reasoning as well as a describing title, preferably using conventional commit semantics.

internal/account/account.go
```diff
@@ -0,0 +1,590 @@
...
+	}
+
+	if a.nauthNamespace == "" {
+		return false
+	}
+
+	return true
+}
+
+func (a *Manager) CreateAccount(ctx context.Context, state *natsv1alpha1.Account) error {
```
kejne: **suggestion:** same here as with manager
```go
account.CreateAccount(...)

// All endpoints
account.Create(...)
account.Delete(...)
account.Import(...)
```

PR #87, opened 2025-12-12
feat: enforce sign-offs of DCO instead of statement in PR

kejne:
The previous statement in the PR to be replaced by a DCO sign-off.
Merging PR:s without it will not be possible with this change.

### DIFF
--- a/.github/actions/require-signed-off/action.yml
+++ b/.github/actions/require-signed-off/action.yml
@@ -4,13 +4,13 @@ description: Ensure every commit in the range carries at least one Signed-off-by
 
 inputs:
   base_sha:
-    description: 'Base commit SHA to compare against (defaults to the pull request base)'
+    description: "Base commit SHA to compare against (defaults to the pull request base)"
     required: false
-    default: ''
+    default: ""
   head_sha:
-    description: 'Head commit SHA to compare (defaults to the workflow SHA)'
+    description: "Head commit SHA to compare (defaults to the workflow SHA)"
     required: false
-    default: ''
+    default: ""
 
 runs:
   using: composite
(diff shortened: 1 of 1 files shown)


.github/actions/require-signed-off/action.yml
```diff
@@ -0,0 +1,61 @@
...
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        BASE_SHA=${{ inputs.base_sha != '' && inputs.base_sha || github.event.pull_request.base.sha }}
+        HEAD_SHA=${{ inputs.head_sha != '' && inputs.head_sha || github.sha }}
+
+        if [[ -z "$BASE_SHA" || -z "$HEAD_SHA" ]]; then
+          echo "::warning::Base or head SHA unavailable; skipping signed-off check."
+          exit 0
```
thobiaskarlsson: **suggestion**: If we should really enforce signed-off commits, exit with error code instead if this ever happens as something is clearly erroneous in the script or the branch tree.

.github/workflows/require-signed-off.yml
```diff
@@ -0,0 +1,20 @@
+name: Require Signed-Off Commits
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
thobiaskarlsson: **question**: Requiring signed-off commits in the PR is one thing, but since we allow Squash-commits, do we enforce this on the final commit that goes to `main`?

thobiaskarlsson (commented):
Great improvement!

.github/actions/require-signed-off/action.yml
kejne: This is basically in the event where you have a brand new repo/initial commit scenario. In that case you don't want the action to fail. So it doesn't have to be incorrect.
But I would agree that it's better to fail in this case.

.github/workflows/require-signed-off.yml
kejne: The sign-off will be added to the squash commit by default. Will look like this:
https://github.com/WirelessCar/nauth/commit/a88d5c9eb5bc014cba66a27c95d4db9d5c0c40c4
(See PR link for ref)

Issue #79, opened 2025-12-03
extract api model from domain


thobiaskarlsson:
**Is your feature request related to a problem? Please describe.**
Currently the services (`AccountManager`/`UserManager`) relies heavily on the `v1alpha1` types. This binds the core business logic to the controller & API layer. The `nauthv1alpha1.Account`/-`.User` parameters passed to the services are mutated by the services and then updated as-is by the controllers. This makes unit testing the adapter (controller) alone fragile as the service mocks needs to behave as the implementations by setting e.g. resulting state on the passed parameters.
 
**Describe the solution you'd like**
To detach the API model from the domain model simplifies testing and enables us to do changes to the API that does not affect the service business logic.

PR #90, opened 2025-12-18
refactor(account): return immutable result instead of mutating state

choufraise:
Change AccountManager interface methods (Create, Update, Import) to return AccountResult instead of mutating the input Account resource. This improves separation of concerns between the manager (computes values) and controller (applies values to Kubernetes resources).

Issue: https://github.com/WirelessCar/nauth/issues/79

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -22,16 +22,11 @@ jq = "latest"
 [settings]
 experimental = true
 
-[tasks.go-test-ginkgo]
-description = "Run Ginkgo tests"
+[tasks.go-test]
+description = "Run go tests"
 dir = "{{ cwd }}"
-alias = "gtg"
-run = "ginkgo run"
-
-[tasks.go-test-ginkgo-all]
-description = "Run all Ginkgo tests in project"
-alias = "gtga"
-run = "ginkgo run -r"
+alias = "gt"
+run = "go test ./..."
 
 [tasks."nauth:e2e-test"]
 silent = true


internal/account/account.go
```diff
@@ -249,104 +248,108 @@ func (a *Manager) Update(ctx context.Context, state *natsv1alpha1.Account) error
...
-
-	return nil
+		return nil, fmt.Errorf("failed to upload account jwt: %w", err)
+	}
+
+	// Return immutable result - controller will apply to state
+	return &AccountResult{
+		AccountID:       accountID,
+		AccountSignedBy: operatorSigningPublicKey,
+		Claims: &natsv1alpha1.AccountClaims{
```
henriropp: Should this not also do `claims := convertNatsAccountClaims(accountClaims)`

internal/account/account.go
```diff
@@ -249,104 +248,108 @@ func (a *Manager) Update(ctx context.Context, state *natsv1alpha1.Account) error
...
 	}
 	accountClaims, err := jwt.DecodeAccountClaims(accountJWT)
 	if err != nil {
-		return fmt.Errorf("failed to decode account jwt for account %s: %w", accountID, err)
+		return nil, fmt.Errorf("failed to decode account jwt for account %s: %w", accountID, err)
 	}
 
-	state.GetLabels()[k8s.LabelAccountSignedBy] = operatorSigningPublicKey
-	state.Status.Claims = convertNatsAccountClaims(accountClaims)
-	state.Status.SigningKey.Name = accountSigningPublicKey
```
thobiaskarlsson: **note**: The SigningKey is no longer set on the resource `Status` as it's not used nor needed. It's kept in the CRD since it might be needed when we implement support for signing key rotation.

internal/controller/account_test.go
```diff
@@ -120,7 +120,12 @@ var _ = Describe("Account Controller", func() {
 			It("should successfully reconcile the account", func() {
 				By("Reconciling the created account")
 
-				accountManagerMock.On("Create", mock.Anything, mock.Anything).Return(nil).Once()
+				mockResult := &AccountResult{
```
thobiaskarlsson: **note**: Mock results have been moved out of the mock to improve test readability. Mocks should be dumb and avoid magic and assumptions.

internal/account/claims_test.go
```diff
@@ -0,0 +1,174 @@
...
+			rebuilder := newClaimsBuilder(ctx, *rebuiltNatsClaims, testClaimsAccountPubKey, accountGetterMock)
+			rebuilder.signingKey(testClaimsSigningKey01)
+			rebuilder.signingKey(testClaimsSigningKey02)
+
+			natsClaimsRebuilt, err := builder.build()
+			require.NoError(t, err)
+			require.NotNil(t, natsClaimsRebuilt)
+			// natsClaims.Encode(operatorSigningKey)
+
+			normalizedNatsClaimsRebuilt := normalizeClaimsForApproval(natsClaims)
```
thobiaskarlsson: **issue**: This is a test bug. We should use `natsClaimsRebuilt` here, not `natsClaims`. Will fix this!

kejne (approved):
Even better! 🎉

mise.toml
```diff
@@ -22,13 +22,8 @@ golangci-lint = "latest"
...
-alias = "gtg"
-run = "ginkgo run"
-
-[tasks.go-test-ginkgo-all]
-description = "Run all Ginkgo tests in project"
-alias = "gtga"
-run = "ginkgo run -r"
\ No newline at end of file
+alias = "gt"
+run = "go test ./..."
```
henriropp: aha, so this runs the ginko tests also?

Issue #74, opened 2025-12-01
Improve default security settings in manifests


kejne:
NAuth does have any requirements of using elevated access or using the local filesystem. For example:

```
securityContext:
  readOnlyRootFilesystem: true
  ...
```

PR #97, opened 2026-01-07
build: read-only root file system

choufraise:
Setting the `readOnlyRootFilesystem` flag to `true` reduces the attack surface of the container since an attacker cannot manipulate the executables of the container, nor make any other changes to the root file system.

Issue: https://github.com/WirelessCar/nauth/issues/74


kejne (approved):

henriropp (approved):
LGTM

PR #99, opened 2026-01-08
feat: stricter user claims conversion

choufraise:
The user claims stated in the k8s resource status shouldn't be a direct copy of the incoming spec, it should represent the actual claims defined in NATS.

### DIFF
--- a/internal/user/approvals/claims_test.TestClaims.defaults.input.yaml
+++ b/internal/user/approvals/claims_test.TestClaims.defaults.input.yaml
@@ -1,0 +1,1 @@
+# No limits specified - should default most to e.g. -1 (unlimited)
--- a/internal/user/approvals/claims_test.TestClaims.defaults.output.nats.approved.json
+++ b/internal/user/approvals/claims_test.TestClaims.defaults.output.nats.approved.json
@@ -1,0 +1,16 @@
+{
+  "iat": 1700000000,
+  "iss": "AAGKQ2YCJCKATG4XUJCLHBSJGNB5TCTJD6N7G6YTKRLOAMGAMQHOZ73L",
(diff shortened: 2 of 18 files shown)


internal/account/claims_test.go
```diff
@@ -96,11 +109,16 @@ func TestClaims(t *testing.T) {
 			rebuilder.signingKey(testClaimsSigningKey01)
 			rebuilder.signingKey(testClaimsSigningKey02)
 
-			natsClaimsRebuilt, err := builder.build()
+			natsClaimsRebuilt, err := rebuilder.build()
```
thobiaskarlsson: **note**: This was the bug that was fixed. Before we compared the same objects equality, which of course will always pass.

internal/account/claims_test.go
```diff
@@ -46,10 +47,21 @@ func TestClaims(t *testing.T) {
...
 			accountGetterMock := NewAccountGetterMock()
 			getAccountCall := accountGetterMock.On("Get", mock.Anything, mock.Anything, mock.Anything)
 			getAccountCall.RunFn = func(args mock.Arguments) {
-				accountID := fakeAccountId(args.String(1), args.String(2))
+				accountRefName := args.String(1)
+				namespace := args.String(2)
+				var accountID string
+				if accountRefName != "" && namespace != "" {
+					accountID = fakeAccountId(accountRefName, namespace)
+				} else if rebuildJwtMode {
```
henriropp: This `rebuildJwtMode` flag looks ugly, "global" flag changing behaviour of function. Is there some other way to do it?

thobiaskarlsson: It's changing the behavior of the mock only to be able to compare the input with the re-built input from the full roundtrip. I'll look into cleaning it up a bit and make it more clear that it's for the rebuilding/final verification only.

PR #104, opened 2026-01-11
feat: use scoped signing keys for users

BZValoche:
Using scoped signing keys allow to:
- keep the permissions inside the NATS resolver
- provide user creds whose permissions can be updated without the need to emit new creds
- revoke the user creds by simply removing the signing key from the NATS resolver

This PR adds a property in the User specification, which can be set to enable the use of signing keys. If not, it works the same than before.

NATS documentation on scoper signing keys: https://docs.nats.io/using-nats/nats-tools/nsc/signing_keys

I might also have corrected some bits of code which do not relate directly with the purpose of this PR, which prevented me to run the operator locally.

Still WIP, there is room for refactor, and I probably missed some cases.

### DIFF
--- a/api/v1alpha1/user_types.go
+++ b/api/v1alpha1/user_types.go
@@ -17,7 +17,10 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"crypto/md5"
+	"encoding/hex"
 	"fmt"
+	"io"
 	"reflect"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -30,6 +33,9 @@ import (
 type UserSpec struct {
 	// AccountName references the account used to create the user.
 	AccountName string `json:"accountName"`
+	// UseSigningKey generates a scopping signing key for the user.
+	// +Optional
+	UseSigningKey bool `json:"useSigningKey,omitempty"`
 	// DisplayName is an optional name for the NATS resource representing the user. May be derived if absent.
 	// +optional
 	DisplayName string `json:"displayName,omitempty"`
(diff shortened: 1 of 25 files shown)


api/v1alpha1/user_types.go
```diff
@@ -96,6 +102,28 @@ func (u *User) GetUserSecretName() string {
 	return fmt.Sprintf("%s-nats-user-creds", u.GetName())
 }
 
+const (
+	SecretNameUserSignTemplate = "%s-u-sign-%s" //  #nosec G101
```
thobiaskarlsson: **suggestion**: This constant is also defined in `internal/user/constants.go`. Suggestion is to keep only one of them.

test/e2e/basic-test/00-assert-operator-secrets.yaml
```diff
@@ -44,7 +44,7 @@ apiVersion: v1
 kind: Secret
 type: Opaque
 metadata:
-  name: operator-sau-creds
+  name:
```
thobiaskarlsson: **question**: Why did you remove the name here? It's a basic test case that (hopefully) should never need to be altered, to verify backward complexity etc.

choufraise (changes requested):
Thank you for providing this. I really like the concept of Scoped Signing Keys, it's a great feature that should be the preferred way of dealing with users (as you mention). However, I have some concerns about this specific implementation of it and see great risks with how it's dealt with. See comments.

choufraise:
@BZValoche , please have a look at issue #140 . Would this solution solve your use case?

BZValoche:
> @BZValoche , please have a look at issue #140 . Would this solution solve your use case?

Yes, definitely. I think you can reject my PR, #140 is way cleaner.

choufraise:
Closing. Planned to be implemented according to #140.

PR #109, opened 2026-01-17
feat(chart): Add .Values.global.labels & normalize all labels

wingsofovnia:
* Adds `.Values.global.labels` for putting custom labels on chart-produced resources. Useful for labelling after env, team or some other policy. Naming is aligned with the [NATS chart `.Values.global.labels`](https://github.com/nats-io/k8s/blob/d0ba0c2df1224f3f0db68d7f8dbc5447a0f066cc/helm/charts/nats/templates/_helpers.tpl#L102-L104) for consistency. Using `global` also helps in case `nauth` is used as a subchart (can be shared across all subcharts, see [doc](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values)) and is a common practice (e.g. cert-manager's [`.Values.global.commonLabels `](https://github.com/cert-manager/cert-manager/blob/9668922e1d443c985864854659f2ab5aa7322c68/deploy/charts/cert-manager/templates/_helpers.tpl#L159C8-L161))
* Cover new variable with tests
* Make all resources to use the same template for labels

Other:
* Re-run `cd charts/ && helm-docs .`
* Add missing `README.md.gotmpl` for `nauth-crds`

### DIFF
--- a/charts/nauth/templates/rbac_account_roles.yaml
+++ b/charts/nauth/templates/rbac_account_roles.yaml
@@ -2,9 +2,9 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: {{ if not .Values.namespaced }}Cluster{{ end }}Role
 metadata:
-  labels:
-    app.kubernetes.io/name: {{ include "nauth.name" . }}
   name: {{ include "nauth.fullname" . }}-account-admin
+  labels:
+    {{- include "nauth.labels" . | nindent 4 }}
   {{- if .Values.namespaced }}
   namespace: {{ include "nauth.namespaceName" . }}
   {{- end }}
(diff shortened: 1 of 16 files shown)


charts/nauth/templates/rbac_account_roles.yaml
```diff
@@ -2,9 +2,9 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: {{ if not .Values.namespaced }}Cluster{{ end }}Role
 metadata:
-  labels:
-    app.kubernetes.io/name: {{ include "nauth.name" . }}
```
wingsofovnia: fyi: this is added by `include "nauth.selectorLabels"` used in `include "nauth.labels"`
```
>  helm template . -s templates/rbac_account_roles.yaml | yq 'select(di == 1) | .metadata.labels'
app.kubernetes.io/name: nauth
app.kubernetes.io/instance: release-name
app.kubernetes.io/version: "0.5.1"
app.kubernetes.io/managed-by: Helm
```

kejne (commented):
Looks good!
As you probably noticed, the commits requires sign-offs now (and no text in the PR):
https://github.com/WirelessCar/nauth?tab=contributing-ov-file#developers-certificate-of-origin

Just needs a quick refactor to make it clearer.

PR #116, opened 2026-01-27
feat: Make `nauth` support multiple NatsClusters

aleksanderaleksic:
## Summary

Introduces the `NatsCluster` CRD to provide a more declarative and extensible approach to configuring NATS authentication backends, replacing the implicit label-based secret discovery mechanism.

**Key changes:**
- New `NatsCluster` CRD (`nauth.io/v1alpha1`) for explicit operator credentials configuration
- Accounts can now reference a `Cluster` via `spec.natsClusterRef` for credential resolution
- Introduces a Provider/Resolver pattern enabling future support for third-party NATS backends
- Full backward compatibility maintained - existing deployments without `natsClusterRef` continue to work

## Motivation

The current nauth implementation requires secrets to be labeled with `nauth.io/secret-type: operator-sign` and `nauth.io/secret-type: system-account-user-creds` for the controller to discover operator credentials. This approach has limitations:

1. **Implicit configuration** - The relationship between accounts and credentials is not explicit in the CRD
2. **Single cluster only** - Only one NATS cluster can be managed per namespace
3. **No extensibility** - Adding support for managed NATS providers (e.g., Synadia Cloud) would require significant rework

The new `NatsCluster` CRD addresses these issues by:
- Making credential references **explicit and declarative** via `natsClusterRef` fields
- Allowing **multiple clusters** to coexist (each account references its cluster)
- Introducing a **Provider interface** that can be implemented for different backends

## New CRD Schema

```yaml
apiVersion: nauth.io/v1alpha1
kind: NatsCluster
metadata:
  name: my-nats-system
  namespace: nauth
spec:
  url: nats://nats.nats.svc.cluster.local:4222
  operatorSigningKeySecretRef:
    name: operator-signing-key
    key: seed
  systemAccountUserCredsSecretRef:
    name: system-account-creds
    key: user.creds
```

Accounts reference a NatsCluster via `natsClusterRef`:

```yaml
apiVersion: nauth.io/v1alpha1
kind: Account
metadata:
  name: my-account
  namespace: nauth
spec:
  natsClusterRef:
    kind: NatsCluster
    name: my-nats-cluster
    namespace: nauth
  # ... rest of account spec
```

## Architecture

```mermaid
flowchart TD
    AR[AccountReconciler] -->|resolves| R[Resolver]
    AR -->|delegates| P[Provider\ninterface]
    R -->|creates| F[Factory]
    F -->|creates| P
    P -.->|implements| NP[nauth.Provider]
    NP -->|wraps| AM[account.Manager]
    NP -->|wraps| UM[user.Manager]
```

- **Resolver**: Determines which Provider to use based on `Account.Spec.natsClusterRef`
- **Provider**: Interface for account/user operations (`CreateAccount`, `CreateUser`, etc.)
- **Factory**: Creates Provider instances configured with the appropriate Cluster

## Backward Compatibility

- Accounts **without** `natsClusterRef` continue to use the legacy label-based secret discovery
- Existing deployments require **no changes**
- The legacy approach may be deprecated in a future release

## Future Extensibility

This architecture enables adding new cluster backends. For example, Synadia Cloud support could be added as:

```yaml
apiVersion: synadia.nauth.io/v1alpha1
kind: NatsCluster
metadata:
  name: ngs-system
spec:
  teamId: "team-id"
  credentialsSecretRef:
    name: synadia-credentials
    key: token
```

Each backend would implement the `cluster.Provider` interface with its specific logic.

## Test Plan

- [x] Unit tests pass (`make test`)
- [x] E2E tests pass (`make test-e2e`)
- [x] New E2E test suite (`cluster-ref-test`) validates the Cluster approach:
  - Creates Cluster with secretRefs (no legacy labels)
  - Creates Account with natsClusterRef
  - Verifies no fallback to legacy label-based discovery
  - Creates User and verifies credentials


kejne:
Hi! Thanks for the PR! Sorry for slow feedback, we'll have a look at the issue and PR to get an understanding since it's a quite big update.
Looks like you got most parts covered test-wise etc :ok_hand: 
Most importantly to see that the concept does not clash and has a good migration path for current users.

internal/controller/account.go
```diff
@@ -136,7 +133,11 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 		if controllerutil.ContainsFinalizer(natsAccount, controllerAccountFinalizer) {
 			if managementPolicy != k8s.LabelManagementPolicyObserveValue {
-				if err := r.accountManager.Delete(ctx, natsAccount); err != nil {
+				provider, err := r.resolver.ResolveForAccount(ctx, natsAccount)
+				if err != nil {
+					return r.reporter.error(ctx, natsAccount, fmt.Errorf("failed to resolve system provider: %w", err))
```
thobiaskarlsson: **suggestion**: Account deletion will be blocked if the referenced System resource is removed first. It's all accurate, as the Account JWT will still be present in the NATS server. Suggestion is to improve this error message to make this clearer to ops, i.e. that the Account is marked for deletion, but that the Account JWT deletion operation cannot be done due to lack of (system) credentials.

internal/system/provider.go
```diff
@@ -0,0 +1,59 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package system
+
+import (
+	"context"
+
+	natsv1alpha1 "github.com/WirelessCar/nauth/api/v1alpha1"
```
thobiaskarlsson: **nit**: I've seen this `natsv1alpha1` alias in other Go files within this repo, but I strongly believe the naming is bad as it's _not_ "NATS API". Consider using e.g. simply `v1alpha1` or `nauthv1alpha1` instead for new files.

cmd/main.go
```diff
@@ -226,29 +225,28 @@ func main() {
...
-	accountManager := account.NewManager(
-		accountClient,
-		natsClient,
-		secretClient,
-	)
+	// Create the system resolver with the nauth provider factory
+	// The factory will create managers with System CRD support when needed
+	nauthFactory := nauth.NewFactory(accountClient, secretClient, natsURL, namespace)
+	resolver := system.NewResolver(mgr.GetClient(), namespace)
+	resolver.RegisterFactory(system.APIGroupNauth, nauthFactory)
```
thobiaskarlsson: **question**: Out of curiosity. With the introduction of resolvers, are you planning on contributing with e.g. a Synadia Cloud provider soon?

internal/system/resolver.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// ResolveForAccount returns the appropriate Provider for the given account
+func (r *DefaultResolver) ResolveForAccount(ctx context.Context, account *natsv1alpha1.Account) (Provider, error) {
+	// Determine API group
+	apiGroup := APIGroupNauth
+	var system *natsv1alpha1.System
+
+	if account.Spec.SystemRef != nil {
+		systemRef := account.Spec.SystemRef
```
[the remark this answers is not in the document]
aleksanderaleksic: I think removing the `apiGroup` in favor of `apiVersion`.
That is more consistent with the `apiVersion` of the resources themself. 

So switch up the check and default to `nauth.io/v1alpha1`

thobiaskarlsson (changes requested):
Looks much better! The tests are not green though, and I have a couple of issues/concerns that you'll have to take a look at. Please do not self-resolve the comments/conversations, I'd appreciate it if you would simply react/respond to them so that I can resolve them when I'm happy with the outcome.

api/v1alpha1/natscluster_types.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// Key in the Secret, when not specified an implementation-specific default key is used.
+	// +optional
+	Key string `json:"key,omitempty"`
+}
+
+// NatsClusterSpec defines the desired state of NatsCluster
+type NatsClusterSpec struct {
+	// URL is the NATS server URL for this cluster.
+	// +required
```
thobiaskarlsson: **question**: In my [question from yesterday regarding the URL](https://github.com/WirelessCar/nauth/pull/116#discussion_r2787923124), it was merely a question as I wasn't sure if it should be in the Cluster CRD or not. I believe it makes sense to _support_ it here, but should it really be _required_? 🤔  Please respond with your thoughts on this as you've thought more about this feature than any one else.

thobiaskarlsson: Awesome! Thank you for this, and great idea with the `urlFrom`. 👌 I leave this decision up to you, if you have time and energy to look into it now, then please go ahead. If not, could I ask you to create an improvement issue with your thoughts on this that anyone can pick up when suited? I'll approve this PR now anyway, please tell me when you want it merged.

thobiaskarlsson (commented):
There are a lot of commits in this PR, most are not following conventional commits format. 

Please update the commit messages or squash them all into one commit so that we can rebase merge (and keep your signature) instead of squash merging.

internal/cluster/nauth/factory.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Create managers with the appropriate configuration
+	accountManager := account.NewManager(f.accounts, natsClient, f.secretClient, opts...)
+	userManager := user.NewManager(f.accounts, f.secretClient)
+
+	return NewProvider(accountManager, userManager), nil
+}
+
+// resolveNatsURL returns the NATS URL from NatsCluster spec (url or urlFrom).
+func (f *Factory) resolveNatsURL(ctx context.Context, nc *nauthv1alpha1.NatsCluster) (string, error) {
```
thobiaskarlsson: **praise**: Nice addition! ❤️

PR #124, opened 2026-02-09
feat(chart): pod teamplate to use nauth.labels

wingsofovnia:
A followup to https://github.com/WirelessCar/nauth/pull/109/ where I forgot to update pod template spec
🙏 I would be enormously grateful if this could be cut into v0.5.3.

### DIFF
--- a/charts/nauth/templates/deployment.yaml
+++ b/charts/nauth/templates/deployment.yaml
@@ -22,7 +22,7 @@ spec:
       {{- end }}
       labels:
         control-plane: controller-manager
-        app.kubernetes.io/name: {{ include "nauth.name" . }}
+        {{- include "nauth.labels" . | nindent 8 }}
     spec:
       serviceAccountName: {{ include "nauth.serviceAccountName" . }}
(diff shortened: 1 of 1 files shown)


charts/nauth/templates/deployment.yaml
```diff
@@ -22,7 +22,7 @@ spec:
       {{- end }}
       labels:
         control-plane: controller-manager
-        app.kubernetes.io/name: {{ include "nauth.name" . }}
+        {{- include "nauth.labels" . | nindent 4 }}
```
wingsofovnia: fyi: `app.kubernetes.io/name` is already included with `nauth.labels` via `nauth.selectorLabels`

thobiaskarlsson (approved):
LGTM. 👍

thobiaskarlsson (commented):
Please ensure the commit is signed-off. 

EDIT: The second commit message is not compliant with Conventional Commits. Feel free to force-push the first commit (incl. sign-off) instead of adding a second one. We do rebase merge so that your commit stays completely verified on main as well, hence this is important.

Issue #73, opened 2025-12-01
Improve release workflow and versioning


kejne:
The release workflow is currently a mix of `.image_version` and `Chart.yaml` versions set.

This allows for setting different versions for different artifacts, but it also becomes confusing since a release is done with a specific version tag.

## Proposal
Remove any versions in files and instead use the git tag as the universal version across charts and images. This will remove any confusion in regards to versioning from a release perspective.

We will release versions where an image or chart hasn't been updated at all, but this would be worth it to keep aligned versioning.

PR #133, opened 2026-02-16
feat(release): automatic versioning based on release tag

choufraise:
Issue: #73


.github/workflows/release.yaml
```diff
@@ -35,10 +35,12 @@ jobs:
           username: ${{github.actor}}
           password: ${{secrets.GITHUB_TOKEN}}
 
-      - name: Read version from .image_version
-        id: nauth
+      - name: Validate semver syntax
+        run: echo "${{ github.ref_name }}" | grep -e "^v[[:digit:]]\{1,3\}\.[[:digit:]]\{1,3\}\.[[:digit:]]\{1,3\}$"
```
cmoscofian: **note**: I don't think this will be an issue, but it's just important to keep in mind that this only works if we ensure that none of our versioning go over 999 on any level.
Maybe by the nature of our cycles might not be an issue, but it's not uncommon for very active projects to go on patch/ minor over 1000.

.github/workflows/release.yaml
```diff
@@ -87,12 +89,18 @@ jobs:
       - uses: actions/checkout@v6
       - uses: azure/setup-helm@v4
       - name: Validate semver syntax
-        run: echo "${{ github.ref_name }}" | grep -e "^v[[:digit:]]\{1,3\}.[[:digit:]]\{1,3\}.[[:digit:]]\{1,3\}$"
+        run: echo "${{ github.ref_name }}" | grep -e "^v[[:digit:]]\{1,3\}\.[[:digit:]]\{1,3\}\.[[:digit:]]\{1,3\}$"
```
cmoscofian: **note**: same as stated before

Issue #134, opened 2026-02-16
Rely on Kind instead of APIVersion to differentiate between cluster types


choufraise:
## Background
In #116 the `NatsClusterRef`- and `NatsCluster` CRD was introduced. The PR also prepared for supporting different types of cluster backends, e.g. Synadia Cloud (see "Future Extensibility" in #116 description). The idea was to differentiate between cluster types based on the `APIVersion`. After further consideration it became clear that we should then differentiate on `Kind` instead of _version_, as e.g. a new `XCloudCluster` CRD would still be bound to the NAuth API as we don't want external CRD dependencies. 

_Note: If this issue is not solved before version 0.5.3+ is released we need to consider this a bug instead of simple refactoring._ 

 ## Todo
* Refactor the provider routing to be based on `Kind` instead of `APIVersion`.
* As we don't have a second cluster type yet, we should fail reconciliation if `Kind` != `NatsCluster` as we need to refactor the factory lookup ([r.factories[...]](https://github.com/WirelessCar/nauth/blob/bc0175d7ca5b4a43085d98d73c2ab1c60aaba266/internal/cluster/resolver.go#L93)) anyway since the [ProviderFactory](https://github.com/WirelessCar/nauth/blob/bc0175d7ca5b4a43085d98d73c2ab1c60aaba266/internal/cluster/resolver.go#L37) is bound to the `v1alpha1.NatsCluster` type only.

aleksanderaleksic:
I guess adding `APIVersion` was a bit premature since we only have one group and one version for the `NatsCluster` kind.
Also I did not really think much of it, I am kind of inspired by the [flux](https://fluxcd.io/flux/components/kustomize/api/v1/#kustomize.toolkit.fluxcd.io/v1.CrossNamespaceSourceReference) interface

#128 that I have drafted introduces `System` through the `synadia.nauth.io` group, so routing based on kind is ok in this scenario. I though that would make sense to clearly separate synadia specifics (systems) from nauth specifics (NatsClusters). 

If you remove API version from the `natsClusterRef` and introduce a new version for the same Kind you would have to fetch the `NatsCluster` by name and read the apiVersion to know what version it is.

To conclude:
I dont see a problem right now routing based on kind, I am wondering what the impact that will have in the future.

---

[...]

Issue #108, opened 2026-01-16
Support for synadia cloud


aleksanderaleksic:
**Is your feature request related to a problem? Please describe.**
As a user of synadia cloud, we want a kubernetes native way of managing accounts, users and credentials.
The problem is that with synadia cloud we dont have access to "system account user credentials" so we are not able to manage accounts or users using `nauth`

**Describe the solution you'd like**
I would like `nauth` integration support with third party NATS providers, for example synadia cloud.

I am imagining introducing something like a "system" kind that we can configure the "system" and reference it in the account spec like in the examples below.
```
---
apiVersion: nauth.io/v1alpha1
kind: System
metadata:
  name: synadia-cloud-local
spec:
  type: synadia_cloud
  synadiaCloud:
     saToken:
        secretRef:
           name: synadia-cloud-sa-token
     systemSelector:
        name: NGS
---
apiVersion: nauth.io/v1alpha1
kind: Account
metadata:
  name: my-acc
spec:
  systemRef:
    name: synadia-cloud-ngs
```

```
---
apiVersion: nauth.io/v1alpha1
kind: System
metadata:
  name: nats-native
spec:
  type: nats-native
  natsNative:
    host: nats.nats.svc.cluster.local
    operatorSecretRef:
        name: nats-operator
---
apiVersion: nauth.io/v1alpha1
kind: Account
metadata:
  name: my-acc
spec:
  systemRef:
    name: nats-native
```

I imagine having a generic "system" interface that supports:
* CRUD operations on accounts
* CRUD operations on users 

**Describe alternatives you've considered**
* Clickops is the current simplest approach in synadia cloud, but it does not scale as the JWTs needs to be copied every time you edit permissions, making it cumbersome and error prone.
* Scripting could also and use the `synctl` cli tool but its not as nice as having it as resources in Kubernetes.
It aslo does not integrate with Kubernetes with secret creation for example.

**Additional context**
I am also questioning if you see such a integration aligns with the goal of this project.

aleksanderaleksic:
Will work on a PR to support this soon given that #116 is merged

aleksanderaleksic:
I guess as a alternative I can just maintain this functionality through a fork or something.

Thanks for considering it 😊

PR #139, opened 2026-02-24
revert: defer multi-cluster provider abstraction to keep nauth scope focused

choufraise:
NAuth is still in an early phase, and we are intentionally keeping scope focused on maturing the core functionality. We are not adding support for additional cluster types (for example Synadia Cloud) at this stage. Introducing multi-integration abstractions now would add significant CRD/API and reconciliation complexity and slow feature development. This change removes that abstraction for now while retaining the explicit NatsCluster/NatsClusterRef support needed today.

Relates to: #108, #134


kejne (approved):
I see that we will want to go this route sometime further on, but with less of a blast radius.
Will be interesting to add support for multiple clusters once the project is ready for it.

PR #146, opened 2026-02-26
fix: default to nauth namspace for default cluster ref

choufraise:
The default cluster reference, global to the operator, may be set by name only (no namespace). This cluster should be used for all accounts the nauth operator manages, hence possibly cross namespaces, hence the default namespace for this cluster reference should be the nauth operator namespace (not the account namespace).

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"crypto/tls"
 	"flag"
+	"log"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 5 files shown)


henriropp (commented):
Great work! Some minor go convention related comments only.

thobiaskarlsson:
Converted back to draft as we found a lot of possible issues with a _default_ cluster. Will go for an _explicit_ cluster instead and make the ref more strict.

thobiaskarlsson:
Closing PR. Will open a new one based on the new findings.

PR #149, opened 2026-03-02
feat: extract nats cluster config lookup from manager

choufraise:
Today we are resolving the cluster specific details within the Account Manager, -Factory and even the NATS client (implementation). With the introduction of `NatsClusterRef` this quickly turns into spaghetti code. A `NatsCluster` resource defaults to the Account namespace if absent, while legacy secrets lookup (System Account Admin creds & Operator Signing Key) defaults to the NAuth Operator namespace. Due to this it's close to impossible to understand what "default-"/"nauth-"/"operator-"/(simply) namespace relates to. Short answer is that it depends on the context, becoming extremely error-prone. Extracting the lookup of cluster details into a separate, testable port ensures separation of concerns and more focused unit tests, not at least for the (business related) account management flows.


internal/account/account_test.go
```diff
@@ -1,1275 +1,3 @@
...
-		Conditions: []v1.Condition{
-			{
-				Type:   ControllerTypeReady,
-				Status: v1.ConditionTrue,
-			},
-		},
-	}
-	return account
-}
+// FIXME: Re-write all Account Manager test cases
```
thobiaskarlsson: **note**: After refactoring into interfaces the original Account Manager tests are not working at all. To re-write them would be horribly with the current _account_ secrets handling we have nested within this manager. The plan is to rely on the E2E tests for now, extract the Account Secrets management to it's own testable interface + impl to ensure that Account Manager unit tests focuses on the actual business flow instead of a X^n amount of secret lookup permutations.

internal/ports/k8s.go
```diff
@@ -0,0 +1,60 @@
+package ports
```
kejne: - **suggestion:** Move structs from the `ports` files to `models`. Good to keep ports clean. I usually use `in` and `out` to indicate in which

PR #160, opened 2026-03-13
feat: strict NamespacedName handling

choufraise:
In nAuth there are several places where we accept an empty namespace and where we then default to e.g. operator-, account- or other CR namespace. This is error-prone due to the risk of passing around "" namespace too deep before it's noticed. Introducing a domain model for dealing with validated Namespace and NamespacedName enabling stricter interfaces and the ability to fail faster with clear expectations.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/WirelessCar/nauth/internal/domain"
 	"github.com/WirelessCar/nauth/internal/nats"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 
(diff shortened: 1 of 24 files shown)


cmd/main.go
```diff
@@ -265,7 +266,7 @@ func main() {
 		namespace = string(controllerNamespace)
 	}
 
-	accountConfig, err := account.NewConfig(operatorNatsCluster, namespace, defaultNatsURL)
+	accountConfig, err := account.NewConfig(operatorNatsCluster, domain.Namespace(namespace), defaultNatsURL)
```
henriropp: This is type conversion? Should use the constructor instead?

choufraise: There is no constructor for the `Namespace` struct as it's just an alias of `string`, hence this is the default "constructor" since I don't explicitly validate the value upon creation. Otherwise a `domain.NewNamespace(string) (Namespace, error)` would have made sense of course.

Issue #165, opened 2026-03-18
nauth should validate jwt using nats-io validation before encoding


choufraise:
## Background

Today nauth have some minimalistic, local claims validations prior to encoding/signing a JWT. The [NATS jwt library](https://github.com/nats-io/jwt/) has built in support for this via the `Validate(...)` functions. 

Update: In JWT lib [v2.8.1](https://github.com/nats-io/jwt/releases/tag/v2.8.1) a fix was introduced allowing stream import of same subject from different accounts. This is critical for JetStream Flow Control (`$JS.FC.>`) to work properly when sourcing from multiple accounts. Ensure this is validated in nauth.

## To Do

- [x] Bump nats-io jwt to latest
- [x] Ensure `jwt.AccountClaims.Validate(...)` and `jwt.UserClaims.Validate(...)` are run before signing JWT.

PR #166, opened 2026-03-18
feat: utilize nats-io jwt claims validation prior to signing

choufraise:
Instead of doing our own minimalistic claims validation we should rely on the existing validation mechanism provided by the nats-io jwt lib.

Bump nats-io/jwt/v2 to v2.8.1 ensure multi-import of same service from different accounts (applicable in e.g. JetStream Flow Control case (`$JS.FC.>`). Verify using local test.

Closes: #165

### DIFF
--- a/internal/account/claims.go
+++ b/internal/account/claims.go
@@ -186,12 +186,6 @@ func newClaimsBuilder(
 			}
 		}
 		claim.Imports = imports
-
-		err := validateImports(imports)
-		if err != nil {
-			errs = append(errs, err)
-			claim.Imports = nil
-		}
 	}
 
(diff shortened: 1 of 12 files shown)


internal/account/claims.go
```diff
@@ -205,24 +199,6 @@ func (b *claimsBuilder) signingKey(signingKey string) *claimsBuilder {
 	return b
 }
 
-func validateImports(imports jwt.Imports) error {
```
choufraise: **note**: Validation removed in favor of the nats-io/jwt/v2 built-in validation function.

eufriction (approved):
LGTM

Issue #178, opened 2026-03-19
reconcile NatsCluster resources and validate cluster connectivity


choufraise:
## Summary

We need to start reconciling `NatsCluster` resources.

Right now, `NatsCluster` objects are not actively reconciled by the operator, which means they never reach a stable ready state from the operator's perspective. In practice, this leaves them stuck in `Progressing...` in e.g. Argo CD.

## Problem

A `NatsCluster` resource should represent a validated and usable NATS cluster target for NAuth, but today we do not verify that the referenced cluster is actually usable.

At minimum, the operator should verify:

1. that the referenced operator signing key secret exists
2. that the referenced Sys Account User Creds secret exists
3. that the Sys Account User Creds can be used to connect to the NATS cluster using the URL defined in the `NatsCluster` resource

Without this, the resource can exist in Kubernetes while still being invalid from NAuth's perspective.

## Proposed behavior

Add reconciliation for `NatsCluster` resources.

The reconciliation should:

- resolve the two required secrets
- read the NATS URL from the `NatsCluster` resource
- attempt to connect to the target NATS cluster using the Sys Account User Creds
- report readiness/status based on whether validation succeeds

If validation succeeds, the resource should become `Ready`.

If validation fails, the resource should surface a clear status/condition message describing what failed:
- missing operator signing key secret
- missing Sys Account User Creds secret
- invalid NATS URL
- failed NATS connection or authentication

## Why

This is needed so that `NatsCluster` reflects real operability instead of just declared configuration.

It will also prevent Argo CD from showing `NatsCluster` resources as perpetually `Progressing...`, and it gives users direct feedback when a cluster target is misconfigured.

## Acceptance criteria

- `NatsCluster` has a reconciler
- the reconciler validates presence of the required secrets
- the reconciler validates connectivity to the configured NATS URL using the Sys Account User Creds
- the resource exposes a meaningful `Ready` condition
- successful validation marks the resource as ready
- validation failures are surfaced through status/conditions with actionable messages
- Argo CD no longer leaves valid `NatsCluster` resources stuck in `Progressing...`

## Notes

This should be a lightweight validation reconciler, not a controller that manages the external NATS cluster itself.

PR #188, opened 2026-03-20
feat: reconcile NatsCluster resources and validate cluster connectivity

choufraise:
## Summary

Add reconciliation for `NatsCluster` resources so their status reflects whether the referenced cluster is actually usable by NAuth.

## Why

`NatsCluster` resources were not being actively reconciled, which meant they could remain in a perpetual `Progressing` state in Argo CD without any real validation from the operator.

This PR makes `NatsCluster` readiness meaningful by validating that the referenced configuration is complete and that the target cluster can actually be reached with the configured system account credentials.

## What Changed

- added status support for `NatsCluster`, including `Ready` conditions
- prepared the core cluster management code for broader `NatsCluster` responsibilities
- added core validation of `NatsCluster` resources:
  - resolve the NATS URL
  - verify the operator signing key secret exists and is valid
  - verify the system account user creds secret exists and is valid
  - connect to the cluster using the configured system account user creds
  - perform a lightweight `$SYS` ping to verify system-account-level access
- added a `NatsCluster` controller that reconciles the resource and reports the validation result through status
- wired the new reconciler into the manager
- added controller and core test coverage for the new validation and reconciliation path
- updated the e2e cluster-ref test to assert that the `NatsCluster` is reconciled successfully
- fixed manager RBAC so the operator can update `natsclusters/status`

## Notes

The new controller tests are still written with Ginkgo to stay consistent with the existing controller test suite until the remaining controller tests have been migrated to Testify as part of #183.

Closes: #178

### DIFF
--- a/internal/adapter/inbound/controller/account_test.go
+++ b/internal/adapter/inbound/controller/account_test.go
@@ -58,7 +58,7 @@ var _ = Describe("Account Controller", func() {
 		ctx := context.Background()
 
 		BeforeEach(func() {
-			operatorVersion = "0.0-SNAPSHOT"
+			operatorVersion = testOperatorVersion
 			_ = os.Setenv(EnvOperatorVersion, operatorVersion)
 
 			accountManagerMock = &AccountManagerMock{}
--- a/internal/adapter/inbound/controller/constants_test.go
+++ b/internal/adapter/inbound/controller/constants_test.go
@@ -1,0 +1,5 @@
+package controller
+
+const (
+	testOperatorVersion = "0.0-SNAPSHOT"
+)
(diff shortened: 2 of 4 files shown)


api/v1alpha1/natscluster_types.go
```diff
@@ -74,15 +74,39 @@ type NatsClusterSpec struct {
 	SystemAccountUserCredsSecretRef SecretKeyReference `json:"systemAccountUserCredsSecretRef"`
 }
 
+// NatsClusterStatus defines the observed state of NatsCluster.
```
kejne: **note:** Good to regenerate the crd documenentation as well

PR #205, opened 2026-03-26
docs(adr): create ADR for import/export split

henriropp:
## Problem statement

The `Account` CRD is currently used to manage everything related to the account, which includes limits, imports and
exports. If any of these settings are broken, especially imports and exports, it can affect the entire account.

Splitting the import and export CRDs from the Account CRD would allow for better separation of concerns.

## Files

* `0005-split-import-and-export-crds-from-account-crd.md` - The ADR itself
* `worklog/0005/` - Contains notes and ideas during creation of ADR. Not up to date, treat them as notes. 

### DIFF
--- a/adr/0005-extract-import-and-export-crds-from-account-crd.md
+++ b/adr/0005-extract-import-and-export-crds-from-account-crd.md
@@ -1,4 +1,4 @@
-# ADR-5: Split import and export CRDs from Account CRD
+# ADR-5: Extract Import- and Export CRDs from Account CRD
 
 Date: 2026-03-26
 
@@ -11,13 +11,13 @@ Splitting the import and export CRDs from the Account CRD would allow for better
 
 ## Status
 
-In progress
+Accepted
 
 ## Context
 
 There exists several identified issues or improvements related to imports and exports, such as: validation of imports
 and matching exports, import and export using activation tokens, support for grouping of related imports and exports (
-for example, jetstream related API topics), getting rid of need to apply imports and exports in correct order.
+for example, JetStream related API subjects).
 
 All these are concerns of imports and exports, and might benefit from the split as well as make the account resources
 more robust and easier to manage.
@@ -36,7 +36,7 @@ they are important to keep in mind when evaluating the options.
 
 - How to handle partial failures, for example, if one of the imports or exports is invalid, should the entire account
   be left out / invalidated or should the valid imports and exports still be applied?
-- Account JWT must be reconciled before imports and exports, otherwise the account might be left in a "broken" state.
+- Account CRD must be reconciled before imports and exports, otherwise the account might be left in a "broken" state.
 - Changes in imports and exports must be reflected in the account JWT, which means that the account JWT must be
   reconciled after any change in imports and exports.
 - How to present reconciliation state in a good way in tools such as `kubectl`, `k9s` and `argoCD`?
@@ -49,7 +49,7 @@ they are important to keep in mind when evaluating the options.
 
 * Cross-cluster imports/exports
 * import and export activation tokens
-* import and export drift detection and Approval workflows
+* import and export drift detection and approval workflows
 
 ## Options
 
@@ -65,7 +65,6 @@ separate CRD.
 
 Same as Option 1 but with separate CRDs for imports and exports.
 Better for separating import and export specific fields and concerns.
-For example `export` has `TokenReq (bool)` field, while `import` has `Token` field.
 
 ### Option 3 - One import/export per import/export CRD
 
@@ -93,12 +92,12 @@ fragments directly in `Account`.
 ## Decision
 
 * Extract import/export to separate CRDs
+* First release must support what is supported today by inline imports and exports
 * Deprecate the import and export fields in `Account` CRD.
 * Use `Option 4 - Contract-based imports/exports` as a target vision
-* First release must support what is supported today by inline imports and exports
-* `tokenReq` and `revocations` in export will be deferred, as these are not implemented today  
+* `tokenReq` and `revocations` in export will be deferred, as these are not implemented today
 
 ## Consequences
 
-* Import and export CRDs can reconcile by themselves. 
+* Import and export CRDs can reconcile by themselves.
 * Failing import and export CRDs will not block account reconciliation.
--- a/adr/0005-extract-import-and-export-crds-from-account-crd.md
+++ b/adr/0005-extract-import-and-export-crds-from-account-crd.md
@@ -1,7 +1,6 @@
-# ADR-5: Extract Import- and Export CRDs from Account CRD
+# ADR-5: Separate AccountImport and AccountExport CRDs and deprecate inline importing and exporting in Account CRD
 
 Date: 2026-03-26
-
 ## Problem statement
 
 The `Account` CRD is currently used to manage everything related to the account, which includes limits, imports and
@@ -91,10 +90,10 @@ fragments directly in `Account`.
 
 ## Decision
 
-* Extract import/export to separate CRDs
+* Extract import/export to separate CRDs (`option 2`).
 * First release must support what is supported today by inline imports and exports
 * Deprecate the import and export fields in `Account` CRD.
-* Use `Option 4 - Contract-based imports/exports` as a target vision
+* Use `Option 4 - Contract-based imports/exports` as a target vision (not all features might be necessary)
 * `tokenReq` and `revocations` in export will be deferred, as these are not implemented today
 
 ## Consequences
--- a/adr/0005-split-import-and-export-crds-from-account-crd.md
+++ b/adr/0005-split-import-and-export-crds-from-account-crd.md
@@ -1,0 +1,291 @@
+# ADR-5: Split import and export CRDs from Account CRD
+
+Date: 2026-03-26
+
+## Problem statement
+
+The `Account` CRD is currently used to manage everything related to the account, which includes limits, imports and
+exports. If any of these settings are broken, especially imports and exports, it can affect the entire account.
+
+Splitting the import and export CRDs from the Account CRD would allow for better separation of concerns.
+
+## Status
+
+In progress
+
+_Allowed values: In progress/Accepted/Superceeded/Rejected/Obsolete._
+
+## Context
+
+There exists several identified issues or improvements related to imports and exports, such as: validation of imports
+and matching exports, import and export using activation tokens, support for grouping of related imports and exports (
+for example, jetstream related API topics), getting rid of need to apply imports and exports in correct order.
+
+All these are concerns of imports and exports, and might benefit from the split as well as make the account resources
+more robust and easier to manage.
+
+It is still the responsibility of the `Account` CRD, to reconcile all the imports and exports that belong to the
+account.
+
+This ADR describes some options for splitting import / export into separate CRDs. It aims to scope the split into
+features that exist today and only introduce new features if they are necessary or small enough. Other features are
+deferred.
+
+### Out of Scope - Deferred to future ADRs
+
+* Cross-cluster imports/exports
+* import and export activation tokens
+* import and export drift detection and Approval workflows
+
+### Concerns
+
+These concerns have been lifted during the writing of the ADR. They are not necessarily all solved by the solutions, but
+they are important to keep in mind when evaluating the options.
+
+- How to handle partial failures, for example, if one of the imports or exports is invalid, should the entire account
+  be left out / invalidated or should the valid imports and exports still be applied?
+- Account JWT must be reconciled before imports and exports, otherwise the account might be left in a "broken" state.
+- Changes in imports and exports must be reflected in the account JWT, which means that the account JWT must be
+  reconciled after any change in imports and exports.
+- How to present the data in a good way in `kubectl` and `k9s`?
+- How to handle overlapping subjects for same target account, from one or more CRs?
+- Can an account in one namespace reference exports/imports in another namespace?
+- Will import/export activation tokens work with these options?
+- Must use deterministic ordering of imports/exports, so that reconciliation does not go into loop.
+
+## Options
+
+### Option 1 - AccountImportExport CRD
+
+Imports and exports are simply moved to a separate CRD, but they are still managed together. The new CRD has a reference
+to the account. This option is in essence the same as the current state, but with imports and exports moved to a
+separate CRD.
+
+```yaml
+# simplified example of the new CRD structure
+AccountImportExport A:
+    account: A
+    imports:
+        - foo.> from Account B
+        - bar.> from Account C
+    exports:
+        - def.> to Account X
+
+AccountImportExport B:
+    account: B
+    exports:
+        - foo.> to Account A
+
+AccountImportExport C:
+    account: C
+    exports:
+        - bar.> to Account A
+
+# alternative for A, can be used if it makes sense to split imports/exports
+AccountImportExport A1:
+    account: A
+    imports:
+        - foo.> from Account B
+
+AccountImportExport A2:
+    account: A
+    imports:
+        - bar.> from Account C
+
+AccountImportExport A3:
+    account: A
+    exports:
+        - def.> to Account X
+```
+
+#### Benefits
+
+- Allows for easiest migration of the existing imports and exports to the new CRD.
+
+#### Drawbacks
+
+- Invites users to define both Export and Import in the same file, even if that's not desired.
+- It's harder to present the data in a good way in `kubectl` and `k9s` as the columns might differ.
+
+### Option 2 - Separate AccountImport and AccountExport CRDs
+
+Same as Option 1 but with separate CRDs for imports and exports.
+Better for separating import and export specific fields and concerns.
+For example `export` has `TokenReq (bool)` field, while `import` has `Token` field.
+
+```yaml
+# simplified example of the new CRD structure
+AccountImport A:
+    account: A
+    imports:
+        - foo.> from Account B
+        - bar.> from Account C
+
+AccountExport A:
+    account: A
+    exports:
+        - def.> to Account X
+```
+
+#### Benefits
+
+- Allows for configuration that affects all listed subjects.
+- Allows for configuration of individual subjects.
+- Less verbose but still allows user to define one import or export per AccountXport if so desired.
+- Allows grouping of related imports and exports, for example, all jetstream related API topics can be grouped together
+  in one CRD.
+    - And as a consequence, it is "all or nothing" for that group, which makes sense when it comes to jetstream.
+
+#### Drawbacks
+
+- Needs to handle partial failures.
+
+### Option 3 - One import/export per import/export CRD
+
+Each import and export is represented as a separate CRD.
+This allows for better granularity and easier management of individual imports and exports,
+but it can lead to a large number of CRDs if there are many imports and exports.
+
+```yaml
+# simplified example of the new CRD structure
+AccountImport A1:
+    account: A
+    subject: foo.>
+    fromAccount: B
+
+AccountImport A2:
+    account: A
+    subject: bar.>
+    fromAccount: C
+
+AccountExport A3:
+    account: A
+    subject: def.>
+    toAccount: X
+```
+
+#### Benefits:
+
+- Very granular validation, i.e. an entire group of rules wouldn't be left out just because one of the subjects are
+  conflicting or invalid.
+
+#### Drawbacks:
+
+- Extremely verbose, at least if we think about import/export for JetStream stuff.
+
+### Option 4 - Contract-based imports/exports
+
+`AccountExport` represents an explicit contract owned by the exporting account.
+`AccountImport` represents an import binding owned by the importing account and references a specific `AccountExport`.
+
+Instead of only defining raw import/export fragments, this option models the relationship between importer and exporter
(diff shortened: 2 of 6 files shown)


eufriction (commented):
Just did a first pass with a few general pointers. Great to start this!

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,291 @@
...
+
+These concerns have been lifted during the writing of the ADR. They are not necessarily all solved by the solutions, but
+they are important to keep in mind when evaluating the options.
+
+- How to handle partial failures, for example, if one of the imports or exports is invalid, should the entire account
+  be left out / invalidated or should the valid imports and exports still be applied?
+- Account JWT must be reconciled before imports and exports, otherwise the account might be left in a "broken" state.
+- Changes in imports and exports must be reflected in the account JWT, which means that the account JWT must be
+  reconciled after any change in imports and exports.
+- How to present the data in a good way in `kubectl` and `k9s`?
```
kejne: Independent on tool used
```suggestion
- How to present the data in a good way in the custom resource?
```

adr/0005-extract-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,291 @@
...
+#### Benefits:
+
+- Very granular validation, i.e. an entire group of rules wouldn't be left out just because one of the subjects are
+  conflicting or invalid.
+
+#### Drawbacks:
+
+- Extremely verbose, at least if we think about import/export for JetStream stuff.
+
+### Option 4 - Contract-based imports/exports
```
kejne, line 74: **question:** I do not really follow what the actual difference is between option 4 and for example option 2? It just looks the same but that option 2 has pseudo config, while option 4 is written to be very exact with how the CRD would look.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,291 @@
...
+
+#### Benefits:
+
+* Imports and Exports can be created in any order.
+* Allows for exporting of group of related subjects, for example, all JetStream API subjects.
+* Open to future improvements such as restricted exports (handling activation tokens) and approval workflows.
+
+#### Drawbacks:
+
+* More complex controller logic to handle the relationships and validation
```
kejne, line 220: **note:** All of the options would require that the account is the one that reconciles, so not sure this is more complex.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,291 @@
...
+    ruleBindings:
+        -   name: "orders eu-west-1"        # human-readable name
+            subject: "orders.eu-west-1.>"   # must match or be a subset of an export rule subject
+            type: service                   # must match the export rule type
+            localSubject: "eu.orders.>"     # optional local subject 
+```
+
+In this example:
+
+* Account A says: "I export these subjects and I allow any account to import them"
```
kejne, line 175: **issue:** There is nothing really bringing up the topic on how to handle private exports in this context.

kejne (commented):
**issue:** I believe that most of what the ADR describes are implementation details. How to handle dependencies between CR:s is something that comes with a controller. But a decision on how the users should interact with the product and the architecture behind it is another.

**suggestion:** Focus simply on the interfaces and if we want to break out the exports/imports at all. Everything underneath is implementation detail (ALL variants of splitting export/imports result in similar need)

There is already a [proposal for the next version](https://github.com/WirelessCar/nauth/blob/main/adr/crds/v1alpha2.md) including the import/export split.

adr/0005-split-import-and-export-crds-from-account-crd.md
kejne: I see that they are present in the worklog, but if you want an ADR then this is something which is good to have here.

henriropp: That was scoped out from this ADR (including using activation tokens). Initially we have the same as today, only public exports. Later we can introduce private exports, but it needs the activation token handling to be secure.

adr/0005-extract-import-and-export-crds-from-account-crd.md
henriropp: The difference is in the concepts. 
In Option 2, the export and import are "standalone". Whatever is put in them will be pushed to NATS, assuming it is valid. 
Option 4 expresses "contracts". Import will not be pushed to NATS unless there is a corresponding Export.

adr/0005-split-import-and-export-crds-from-account-crd.md
henriropp: There is some more validations and checks that an import matches an export.

adr/0005-extract-import-and-export-crds-from-account-crd.md
choufraise: I truly believe that we need to separate those. There is nothing constraining an account to have imports bound to a non-existing export and vice versa. To avoid race conditions and temporary conflicts, where e.g. the exporting account temporarily looses its exports due to a malformed CR or any other reason, we'd still like the already approved imports to not be touched in my opinion. Then _if_ we can visualize that the import is "broken" due to e.g. a lack of export, then that's just a CR Status related feature that shouldn't affect the actual Account JWT.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,116 @@
...
+## Status
+
+In progress
+
+_Allowed values: In progress/Accepted/Superceeded/Rejected/Obsolete._
+
+## Context
+There exists several identified issues or improvements related to imports and exports, such as: validation of imports and matching exports, import and export using activation tokens, support for grouping of related imports and exports (for example, jetstream related API topics), getting rid of need to apply imports and exports in correct order.
+
+All these are concerns of imports and exports, and would benefit from the split as well as make the account resources more robust and easier to manage.
```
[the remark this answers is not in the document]
choufraise: Now this is more considered being a nitpick rather than an actual issue. Focusing on the assumption of "easier to manage" is a minor, neglectable comment. Closing.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,726 @@
...
+```yaml
+AccountExport:
+  access:
+    mode: restricted
+    allowedAccountRefs:
+      - name: Account B
+        namespace: namespace-b
+```
+
+With this model, an `AccountImport` from Account B is not enough on its own. The referenced `AccountExport` must also
```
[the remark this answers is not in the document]
choufraise: In the latest PR, activation tokens are deferred to the future, hence this comment can be ignored.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,726 @@
...
+Because multiple imports and exports may contribute to the final account state, the controller must resolve them
+deterministically.
+
+This includes:
+
+* deterministic ordering of imports and exports
+* conflict detection when subjects overlap
+* clear reporting when two resources cannot be merged safely
+
+For example, if two imports resolve to the same local subject, the controller must report that conflict instead of
```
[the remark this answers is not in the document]
choufraise: Fixed. The text now mentions relying on JWT v2 lib for validation.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,726 @@
...
+            responseType: single
+        -   name: with-extra-options
+            subject: "my.subject.>"
+            type: stream
+            options: # this is optional extra parameters (example struct below)
+                advertise: true
+                allowTrace: true  
+```
+
+`rules[].name` must be unique within the export and is used for reference in the import.
```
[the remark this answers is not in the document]
choufraise: This document is now considered pure work log notes, hence ignoring this issue for the sake of the ADR (which mentions that subject is the main identifier).

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,726 @@
...
+        name: Account B
+        namespace: namespace-b
+    exporterAccountRef:
+        name: Account A
+        namespace: namespace-a
+    exportRef:
+        name: jetstream-api
+        namespace: namespace-a
+    updatePolicy: automatic             # automatic (default) = advance status when export changes compatibly, 
+                                        # manual = detect drift and stop advancing until spec says to accept it
```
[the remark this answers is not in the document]
choufraise: Deferred to future in current PR. Resolved.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,726 @@
...
+    localSubjectPrefix: ext.account-a   # optional prefix to add to all imported subjects, for example, to avoid conflicts with local subjects
+    ruleBindings:   # required, non-empty list
+        -   name: info                  # export rule name
+            expected:                       # validation only fields 
+                subject: "$JS.API.INFO"     # validate remote exported subject
```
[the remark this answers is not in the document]
choufraise: This has been flattened out in latest PR. Resolved.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,726 @@
...
+
+#### Full example
+
+```yaml
+AccountExport:
+    name: jetstream-api
+    namespace: namespace-a
+    accountRef:
+        name: Account A           # expected to be in same namespace
+        namespace: namespace-a
```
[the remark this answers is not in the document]
choufraise: This has been considered in latest changes. Resolved.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,726 @@
...
+    access:
+        mode: restricted       # restricted = only allowed accounts can import, public = any account can import
+        allowedAccountRefs:    # only needed if mode is restricted, otherwise ignored
+        -   name: Account B
+            namespace: namespace-b
```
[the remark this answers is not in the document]
choufraise: Considered in latest PR changes. Resolved due to deferred to future.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,726 @@
...
+```go
+package v1alpha2
+
+type AccountExportRule struct {
+    Name         string       `json:"name,omitempty"`
+    Subject      Subject      `json:"subject,omitempty"`
+    Type         ExportType   `json:"type,omitempty"`
+    ResponseType ResponseType `json:"responseType,omitempty"`
+
+    ExportOptions *ExportOptions `json:"options,omitempty"`
```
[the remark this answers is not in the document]
choufraise: Considered due to the flattening of the structures. Resolved.

adr/0005-split-import-and-export-crds-from-account-crd.md
choufraise: **suggestion**: The concern was not really about making the data look nice but rather ensure binary state for each custom resource, e.g. green/red in visualization tools like kubectl, k9s or ArgoCD.

adr/0005-extract-import-and-export-crds-from-account-crd.md
choufraise: This has been considered in the latest PR changes. The decision is no longer based on implementation details, but rather on the fact to extract import/exports from the Account CRD with target designs/concepts. Resolving.

adr/0005-split-import-and-export-crds-from-account-crd.md
choufraise: I agree with Henri this this option is a bit more complex regarding conciliation than the other options, due to the contractual nature of the relationship between Import and Export. Resolving.

adr/0005-split-import-and-export-crds-from-account-crd.md
choufraise: Private exports deferred from this ADR. Resolving.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,104 @@
+# ADR-5: Split import and export CRDs from Account CRD
```
choufraise:
```suggestion
# ADR-5: Extract Import- and Export CRDs from Account CRD
```

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Status
+
+In progress
+
+## Context
+
+There exists several identified issues or improvements related to imports and exports, such as: validation of imports
+and matching exports, import and export using activation tokens, support for grouping of related imports and exports (
+for example, jetstream related API topics), getting rid of need to apply imports and exports in correct order.
```
choufraise:
```suggestion
for example, JetStream related API subjects).
```
I guess it should be `subjects` instead of `topics`, also there is never a "need to apply imports and exports in correct order". You can define either of them first, they simply just doesn't work until they are matched on both sides.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,104 @@
...
+deferred.
+
+### Concerns
+
+These concerns have been lifted during the writing of the ADR. They are not necessarily all solved by the solutions, but
+they are important to keep in mind when evaluating the options.
+
+- How to handle partial failures, for example, if one of the imports or exports is invalid, should the entire account
+  be left out / invalidated or should the valid imports and exports still be applied?
+- Account JWT must be reconciled before imports and exports, otherwise the account might be left in a "broken" state.
```
choufraise:
```suggestion
- Account CRD must be reconciled before imports and exports, otherwise the account might be left in a "broken" state.
```
Don't we mean "Account CRD" and not "JWT" here?

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,104 @@
...
+- How to handle overlapping subjects for same target account, from one or more CRs?
+- Can an account in one namespace reference exports/imports in another namespace?
+- Will import/export activation tokens work with these options?
+- Must use deterministic ordering of imports/exports, so that reconciliation does not go into loop.
+
+### Out of Scope - Deferred to future ADRs
+
+* Cross-cluster imports/exports
+* import and export activation tokens
+* import and export drift detection and Approval workflows
```
choufraise:
```suggestion
* import and export drift detection and approval workflows
```

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,104 @@
...
+
+Imports and exports are simply moved to a separate CRD, but they are still managed together. The new CRD has a reference
+to the account. This option is in essence the same as the current state, but with imports and exports moved to a
+separate CRD.
+
+### Option 2 - Separate AccountImport and AccountExport CRDs
+
+Same as Option 1 but with separate CRDs for imports and exports.
+Better for separating import and export specific fields and concerns.
+For example `export` has `TokenReq (bool)` field, while `import` has `Token` field.
```
choufraise: **suggestion**: This line feels implicit, I'd remove it entirely.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,104 @@
...
+
+This makes the relationship explicit, easier to validate, and easier to reason about than embedding raw import/export
+fragments directly in `Account`.
+
+## Decision
+
+* Extract import/export to separate CRDs
+* Deprecate the import and export fields in `Account` CRD.
+* Use `Option 4 - Contract-based imports/exports` as a target vision
+* First release must support what is supported today by inline imports and exports
```
choufraise, line 94: **suggestion**: I'd like to see this bullet as the second bullet, before deprecation- and target bullet from an importance- and/or chronological ordering.

adr/worklog/0005/option-4-appendix-reconcile-flow.md
choufraise, line 1: **note**: Explicitly ignoring reviewing this file due to complexity and being just a work log.

adr/worklog/0005/option-4-appendix-resource-relationships.md
choufraise, line 1: **note**: Explicitly ignoring review of this appendix due to AI generated complexity.

adr/worklog/0005/options.md
```diff
@@ -0,0 +1,228 @@
...
+
+To identify conflicts, we will use nats.io's JWT imports validation function.
+
+### Benefits:
+
+* Imports and Exports can be created in any order.
+* Allows for exporting of group of related subjects, for example, all JetStream API subjects.
+* Open to future improvements such as restricted exports (handling activation tokens) and approval workflows.
+
+### Drawbacks:
```
choufraise, line 218: **suggestion**: Just like Option 2, this also requires explicit handling of partial failures. In fact, the only option that this drawback does not apply to is Option 3.

adr/0005-split-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,104 @@
...
+## Problem statement
+
+The `Account` CRD is currently used to manage everything related to the account, which includes limits, imports and
+exports. If any of these settings are broken, especially imports and exports, it can affect the entire account.
+
+Splitting the import and export CRDs from the Account CRD would allow for better separation of concerns.
+
+## Status
+
+In progress
```
choufraise: **suggestion**: Since we have a decision if this PR is merged, then this status should align with this, i.e. this ADR is no longer In progress if merged as-is.

choufraise (commented):
Small comments, then I'll gladly approve! 👍  Nice job!

choufraise (dismissed):
Let's go!

adr/0005-extract-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,103 @@
...
+
+* the exporter publishes what it is willing to expose
+* the importer declares which export it wants to consume
+
+This makes the relationship explicit, easier to validate, and easier to reason about than embedding raw import/export
+fragments directly in `Account`.
+
+## Decision
+
+* Extract import/export to separate CRDs
```
eufriction: suggestion: Based on that Option 4 is the target vision, I guess this is Option 2? It would make sense to include which option is used as a starting point for clarity.

adr/0005-extract-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,103 @@
+# ADR-5: Extract Import- and Export CRDs from Account CRD
```
eufriction: suggestion: make the decision clear in the title, this is very valuable when scanning the entire TOC for what decision have been taken. If it's hidden, every ADR needs to be opened and read.

eufriction (commented):
A few suggestions for clarity, but all in all, I think it looks ready!

adr/0005-extract-import-and-export-crds-from-account-crd.md
henriropp: Naming it `Separate AccountImport and AccountExport CRDs and deprecate inline importing and exporting in Account CRD`

choufraise: I don't agree. We said yesterday that the decision is purely on actually extracting the imports and exports to separate CRDs. The option 4 is merely a sub-decision or the most probably future vision/target. This was the scoping we did yesterday, to not decide on full featured implementation details but rather on the concept itself.

choufraise: ...and [that commit](https://github.com/WirelessCar/nauth/pull/205/commits/aa3543d71e3d8db9eec4a38aef7f502a8560994e)  made sense, great! Ignore my previous comment. 😄

choufraise (approved):

eufriction (approved):

adr/0005-extract-import-and-export-crds-from-account-crd.md
```diff
@@ -0,0 +1,102 @@
...
+This makes the relationship explicit, easier to validate, and easier to reason about than embedding raw import/export
+fragments directly in `Account`.
+
+## Decision
+
+* Extract import/export to separate CRDs (`option 2`).
+* First release must support what is supported today by inline imports and exports
+* Deprecate the import and export fields in `Account` CRD.
+* Use `Option 4 - Contract-based imports/exports` as a target vision (not all features might be necessary)
+* `tokenReq` and `revocations` in export will be deferred, as these are not implemented today
```
kejne, line 97: **nit:** Ideally, the decision should be a one-liner about which option was chosen. However, I believe it is clear enough why and what is decided. It can be for future reference.

kejne (approved):
Looks good!
The options don't really cover how the different concerns are addressed but I believe that the ADR is still on a low enough level that it doesn't need to be perfect. Important is that you have evaluated the approaches and identified the different cases that could arise 👍

PR #206, opened 2026-03-27
feat: reconcile accounts when referenced NatsCluster changes

choufraise:
## DRAFT NOTE
This is a draft PR with a bunch of `FIXME`s that needs to be handled and revised before this can be set to Ready.

## Summary

Reconcile `Account` resources when their related `NatsCluster` changes.

## Why

Accounts depend on cluster-derived state such as the cluster URL, system account credentials, and operator signing key. Those inputs can change even when the `Account` spec does not, so account reconciliation must also react to related `NatsCluster` updates.

## Notes

This change tracks the resolved `NatsCluster` on the account state and labels, and wires the account controller to enqueue affected accounts when a `NatsCluster` changes.

### DIFF
--- a/internal/adapter/inbound/controller/account.go
+++ b/internal/adapter/inbound/controller/account.go
@@ -27,11 +27,14 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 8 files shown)


kejne (commented):
**suggestion:** If we want to link any particular usage, I believe we should use an "Usage" object (similar [as Crossplane does](https://marketplace.upbound.io/providers/crossplane-contrib/provider-aws/v0.37.0/resources/aws.crossplane.io/ProviderConfigUsage/v1beta1)) for this instead. 

This is ofc another CRD, but has a clear purpose. We've already built away labels once before and there is no particular use case where one would gitops this label for example.

PR #212, opened 2026-03-31
test: add decentralized-auth regression tests for NATS JWT assumptions

choufraise:
## Summary

This PR adds embedded NATS integration tests in internal/adapter/outbound/nats/general_test.go to verify decentralized-auth behavior directly against the pinned nats-server and jwt/v2 dependencies.

The goal is to lock down assumptions that NAuth relies on for account exports, imports, and activation-token scoping. If a future dependency bump changes that behavior, these tests should fail early and force a review of the affected NAuth design.

## What Changed

- added a dedicated `nats-regression_test.go` for raw NATS decentralized-auth verification
- starts an embedded NATS server in operator mode
- applies account JWTs after startup through the system claims update API
- signs account JWTs with the operator key
- signs user JWTs with the account root key only
- verifies public export/import behavior across accounts
- verifies token-protected export/import behavior with activation-token scoping

## Why

NAuth depends on security-sensitive behavior in upstream NATS auth and JWT handling.

These tests give us a cheap regression net for cases where:

- nats-server changes claim validation behavior
- jwt/v2 changes import/export or activation-token semantics
- Dependabot bumps either dependency and silently shifts behavior that NAuth assumes today

If one of these tests starts failing after an upstream version change, that is a signal to re-evaluate parts of NAuth rather than blindly accept the bump.

### DIFF
--- a/internal/adapter/outbound/nats/nats_regression_test.go
+++ b/internal/adapter/outbound/nats/nats_regression_test.go
@@ -1,0 +1,315 @@
+package nats
+
+import (
+	"encoding/json"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/nats-io/jwt/v2"
+	natsserver "github.com/nats-io/nats-server/v2/server"
+	"github.com/nats-io/nats.go"
+	"github.com/nats-io/nkeys"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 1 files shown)


kejne (approved):
**praise:** Great! Good direction to use the embedded nats server to make reliable tests!

Issue #185, opened 2026-03-20
remove core/controller dependencies on adapter k8s package


choufraise:
## Problem

`internal/core` and controller code currently depend on (mainly constants from) the adapter `k8s` package.

That gives us the wrong dependency direction and leaks adapter concerns into layers that should not know about infrastructure-specific details.

## Proposal

Identify the constants currently imported from the adapter `k8s` package and move them to the correct layer.

The goal is:

- `core` must not depend on adapter code
- controllers must not depend on unrelated adapter code
- shared concepts should live in a non-adapter package
- adapter-specific constants should remain local to the adapter layer

## Acceptance criteria

- `internal/core` no longer imports adapter `k8s` package
- controller code no longer depends on unrelated adapter packages for shared constants
- shared constants are moved to an appropriate non-adapter package
- adapter-specific constants are only used inside the adapter layer

## Notes

This is a structural refactor only. No intended behavior change.

PR #216, opened 2026-04-10
refactor: move CRD label ownership out of the k8s adapter

choufraise:
## Summary

  This refactor moves Account, User, and management-policy label ownership into the API/core layer and removes the remaining CRD label dependency on the Kubernetes adapter.

  Issue: #185

  ## What changed

  - added shared label types and helpers in `api/v1alpha1` for Account, User, and management policy labels
  - replaced direct `Labels[...]` access with typed getters/setters on `Account` and `User`
  - updated account reconciliation to resolve management policy through the API layer instead of adapter constants
  - moved account secret-specific labels into `internal/core`
  - removed CRD label constants from `internal/adapter/outbound/k8s`, leaving only adapter-specific secret management labels
  - updated core and controller tests to match the new label ownership model
  - removed the leftover kubebuilder scaffolding comment in `user_types.go`

  ## Why this is correct

  These labels describe CRD state and domain behavior. They do not belong to the outbound Kubernetes adapter. Centralizing them in `api/v1alpha1` and `internal/core` removes the core-to-adapter dependency, keeps label handling consistent, and preserves existing behavior with the same label keys.


internal/core/account_claims.go
```diff
@@ -174,7 +173,7 @@ func newAccountClaimsBuilder(
 					accountRef,
 					err))
 			} else {
-				account := importAccount.Labels[k8s.LabelAccountID]
+				account := importAccount.GetAccountID()
```
henriropp: getAccountID()  -- (label hidden)

internal/core/account_claims_test.go
```diff
@@ -46,9 +45,7 @@ func Test_AccountClaims(t *testing.T) {
 			getAccountCall := accountReaderMock.mockGetCallback(mock.Anything, mock.Anything, func(accountRef domain.NamespacedName) (*v1alpha1.Account, error) {
 				accountID := fakeAccountId(accountRef)
 				account := &v1alpha1.Account{}
-				account.Labels = map[string]string{
-					k8s.LabelAccountID: accountID,
-				}
+				account.SetLabel(v1alpha1.AccountLabelAccountID, accountID)
```
henriropp, line 48: setLabel(AccountID) -- (label exposed)

internal/core/user_test.go
```diff
@@ -160,10 +159,10 @@ func (t *UserManagerTestSuite) Test_CreateOrUpdate_ShouldSucceed_WhenUpdatedUser
 	t.NotNil(signedUserJWT, "signedUserJWT not set")
 	t.NotNil(caughtSecrets, "caughtSecrets not set")
 
-	userID := user.GetLabels()[k8s.LabelUserID]
+	userID := user.GetLabelledUserID()
 	t.NotEmpty(userID, "UserID label should not be empty")
-	t.Equal(accountID, user.GetLabels()[k8s.LabelUserAccountID])
-	t.Equal(accountSignPub, user.GetLabels()[k8s.LabelUserSignedBy])
+	t.Equal(accountID, user.GetLabelledAccountID())
```
henriropp: user.get "labelled" AccountID --- (Label hidden, but made aware of as label)

henriropp (dismissed):
I like the idea of hiding the labels into struct methods. 

Currently there is kind of a mix of how it is done. See the comments for examples.
Maybe consistently keep all hidden, getAccountID(), setAccountID(id). 

unless there is a reason to expose and be explicit of the label?

henriropp (approved):

PR #218, opened 2026-04-13
feat: scaffold experimental AccountExport CRD and controller

henriropp:
This is a merge of base scaffolding code for AccountExport CRD. Feature not complete.

## Summary
- add the `AccountExport` API type and generated CRDs
- add an experimental `AccountExport` reconciler with status conditions
- wire the controller behind a new `--experimental-account-export` flag
- expose the feature flag through the Helm chart and add RBAC for `accountexports`
- add an example manifest for the new resource

## Details
This PR introduces the first pass of `AccountExport` support as an opt-in, experimental feature. It adds the resource definition, reconciliation wiring, status reporting, chart/config integration, and example manifests needed to start exercising the API.

## Current limitations
- `AccountExportManager.CreateClaim(...)` is still not implemented
- adoption into the parent `Account` status is still TODO
- this PR is scaffolding/enablement work, not full export validation or end-to-end import/export support yet


choufraise (dismissed):

api/v1alpha1/account_export_types.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (a *AccountExport) GetConditions() *[]metav1.Condition {
+	return &a.Status.Conditions
+}
+
+// AccountExportSpec defines the desired state of AccountExport.
+type AccountExportSpec struct {
+	// AccountName refers to the Account in the same namespace to which this export applies.
+	// +required
+	AccountName string `json:"accountName"`
```
kejne, line 47: **note:** I can see a use case for a cluster-scoped account as well. Adding this as an object which only has `name` as required field and future optional `kind` (which could be `ClusterAccount`) could be a good option.

```suggestion
	AccountRef AccountRef `json:"accountRef"`
```

api/v1alpha1/account_export_types.go
```diff
@@ -0,0 +1,115 @@
...
+	// Rules defines the export rules for this account export. Must have at least one rule.
+	// +required
+	// +kubebuilder:validation:MinItems=1
```
kejne, line 50: **praise:** good that you put in the effort of setting the markers properly right away 👍

api/v1alpha1/account_export_types.go
```diff
@@ -0,0 +1,115 @@
...
+	Advertise *bool `json:"advertise,omitempty"`
+	// +optional
+	AllowTrace *bool `json:"allowTrace,omitempty"`
+}
+
+// AccountExportStatus defines the observed state of AccountExport.
+type AccountExportStatus struct {
+	// Normalized claim for account to use
+	// +optional
+	Claim *AccountExportClaim `json:"claim,omitempty"`
```
kejne, line 79: **question:** Do we have an immediate need for this? What normalization is needed which would not also automatically be done when reconciling the account?

charts/nauth/templates/deployment.yaml
```diff
@@ -37,6 +37,8 @@ spec:
             - --health-probe-bind-address=:8081
             - --metrics-bind-address=:8080
             - --metrics-secure=false
+            # TODO: [#11] Remove experimental flag for AccountExport CRD
+            - --experimental-account-export={{ .Values.experimental.accountExport.enabled }}
```
kejne, line 41: **praise:** good to roll it out feature flagged 👍

internal/core/account.go
```diff
@@ -188,6 +188,11 @@ func signAccountJWT(claims *jwt.AccountClaims, operatorSigningKey nkeys.KeyPair)
 	return claims.Encode(operatorSigningKey)
 }
 
+func (a *AccountManager) ValidateExports(ctx context.Context, state *v1alpha1.AccountExport) (*domain.AccountExportClaim, error) {
+	// TODO: [#22] Implement ValidateExports
+	panic("TODO: [#22] Implement ValidateExports")
```
kejne: **issue:** everything we merge to main should be possible to release if we need to hot-fix issues. Ensure that new functionality cannot accidentally break the operator.

kejne (commented):
Looking very good!
I did have some items that could be good to look at, but good direction! 🚀

api/v1alpha1/account_export_types.go
choufraise: Could you please elaborate on this `ClusterAccount` type? 

Right now we've explicitly chosen to only support account `Name` since an `AccountExport` is an extension to the `Account` CRD, hence they should be part of the same namespace. Creating a specific struct for this type of reference feels overkill, at least since we cannot/should not re-use the existing `AccountRef` which already today supports `Namespace`.

api/v1alpha1/account_export_types.go
choufraise: We want to differentiate between the "validated" rules, as Claim, and the spec rules so that the spec could for instance be at generation 5 (and invalid) while the claimed rules could still be bound to generation 4. This was an attempt to ensure that the exports was simply not just instantly removed from the Account JWT if the user would change the rules to something invalid. Calling them "normalized claim" was an attempt to say that the claimed rules structure must not be 1-1 to the spec rules as we could infer rule fields from other sources in the future. This mainly applies to to future `AccountImport` CRD though, where we see the possibility to import rules by e.g. name only and implicitly stating the rest of the import fields from the referenced `AccountExport`.

internal/core/account.go
choufraise: This code is not yet called by the controller, hence just a place holder to ensure that the `core.AccountManager` implements all methods from `inbound.AccountManager` interface.

choufraise: But sure, we shouldn't panic, will change to a regular error instead.

api/v1alpha1/account_export_types.go
kejne: It is a convention that you append `Cluster` to indicate that a resource is cluster-scoped in the case where you have both namespaced and cluster scoped.
I just foresee that we could very well see the concept of a cluster-scoped account at some point. 
The `AccountRef` was just an example name and it was coincidental that I chose the same name :slightly_smiling_face: 

That being said, maybe it's better to keep to YAGNI since it would be trivial to add a reference at a later stage if wanted.

api/v1alpha1/account_export_types.go
kejne: That makes perfect sense! :ok_hand:

kejne (approved):
LGTM!

api/v1alpha1/account_export_types.go
choufraise: If we would ever introduced cluster-scoped accounts, let's fix it when that heppens.

choufraise (approved):

Issue #11, opened 2025-08-15
Separate Imports and Exports to separate CRDs


kejne:
# Background
An account CR is completely reliant on the entire manifest being valid. However, this can be quite when working with many imports and exports.
It also requires the import and exports to have been made in the correct order.

# Acceptance Criteria
- Break apart the `Account` CRD like it is described in the `v1alpha2` description under `adr`.
- Only valid exports and imports are added to the account jwt

# Not in scope
- Exports are still public exports and NAuth is responsible for validating that the import is allowed.

henriropp:
https://github.com/WirelessCar/nauth/releases/tag/v0.7.0-rc.1

henriropp:
https://github.com/WirelessCar/nauth/releases/tag/v0.7.1

PR #220, opened 2026-04-14
feat: add account adoption status model for aggregated child CRDs

choufraise:
Introduce Account status fields for tracking adoption decisions for child resources such as AccountExport. The API now includes structured adoption entries with resource identity, observed generations, and adoption result
metadata.

This lays the API/schema groundwork for surfacing parent aggregate decisions on child resources without giving the parent reconciler direct ownership of child status.

Issue: #11


api/v1alpha1/account_types.go
```diff
@@ -126,6 +129,47 @@ func (a *Account) SetLabel(label AccountLabel, value string) {
...
+	// +kubebuilder:validation:Minimum=0
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	ClaimObservedGeneration int64 `json:"claimObservedGeneration,omitempty"`
+	// status of the adoption, one of True, False, Unknown.
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=True;False;Unknown
+	Status metav1.ConditionStatus `json:"status"`
```
choufraise, line 161: **note**: The idea is that a child resource reconciler will lookup the `Account`, find it's own `AccountAdoption` based on `UID` and then populate a `AdoptedByAccount` `Condition` in its own `status.conditions` array. The idea is to re-use the `ClaimObservedGeneration`, `Status`, `Reason` and `Message` directly so that both the `Account` and child resource tells the same story.

kejne: I believe that it would be helpful that we start having a docs dir with these flows described for an agent. Just some mermaid diagram and terse text about the major flows.

api/v1alpha1/account_types.go
```diff
@@ -126,6 +129,49 @@ func (a *Account) SetLabel(label AccountLabel, value string) {
...
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	// ClaimObservedGeneration refers to the observed generation of the child resource claim.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	ClaimObservedGeneration int64 `json:"claimObservedGeneration,omitempty"`
+	// Status of the adoption, one of True, False, Unknown.
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=True;False;Unknown
+	Status metav1.ConditionStatus `json:"status"`
```
kejne, line 161: **note:** I might be more clear if the status was a child object in this case - mostly for readability when describing the manifest.

api/v1alpha1/account_types.go
choufraise: I totally agree! We are not 100% sure about if the current reconciliation workflow and relationships are the best choice, so I'd rather document it before feature release (when feature flag is removed). Created a Todo section in #11 and added this as a final todo.

api/v1alpha1/account_types.go
choufraise: Great suggestion! I've added a new commit where I've extracted the status related fields into a child object. Please have a look.

kejne: Looks good!

choufraise:
Awaiting merge of https://github.com/WirelessCar/nauth/pull/221 due to changes in claim naming.

api/v1alpha1/account_types.go
```diff
@@ -126,6 +129,55 @@ func (a *Account) SetLabel(label AccountLabel, value string) {
 	a.Labels[string(label)] = value
 }
 
+// AccountAdoptions defines the status of child resources that have been adopted or are candidates for adoption by this account.
+type AccountAdoptions struct {
+	// Exports defines adoptions of type `AccountExport` that are bound to the account.
+	Exports []AccountAdoption `json:"exports,omitempty"`
```
henriropp, line 135: Do we need validation here? min: 1 ?

api/v1alpha1/account_types.go
```diff
@@ -126,6 +129,55 @@ func (a *Account) SetLabel(label AccountLabel, value string) {
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name,omitempty"`
+	// UID of the child resource UID
+	// +required
+	UID types.UID `json:"uid,omitempty"`
+	// ObservedGeneration refers to the observed generation of the child resource.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
```
henriropp: Shouldn't this be required? If we have UID (which is required), then we should have the generation?

henriropp (dismissed):
Some validation related comments...

api/v1alpha1/account_types.go
choufraise: No, there are no requirements that you have any exports bound to the account.

api/v1alpha1/account_types.go
choufraise: Good point! Will fix this.

choufraise: Added separate commit for this.

henriropp (approved):

PR #221, opened 2026-04-15
feat(account-export): validate export rules before updating desired claim

henriropp:
## Summary

  Validate `AccountExport` rules before exposing them in status so invalid export definitions are rejected early and the controller keeps the last valid desired claim.

  ## What changed

  - implement `AccountExportManager.CreateClaim` using NATS JWT export validation
  - surface invalid export rules through status conditions instead of publishing a new desired claim
  - preserve the last valid desired claim when a later reconcile sees invalid rules
  - rename the status claim field to clarify that it represents the desired normalized claim
  - regenerate CRD manifests and add controller/core test coverage for valid and invalid rule flows

  ## Notes

  - This branch also changes `.mise-tasks/nauth/install.sh` to stop auto-installing examples during local setup. 


.mise-tasks/nauth/install.sh
```diff
@@ -13,5 +13,5 @@ kubectl create namespace "$NATS_NAMESPACE"
 # mise run nauth:install-prometheus
 mise run nauth:install-nats
 mise run nauth:install-nauth
-mise run nauth:install-examples
+#mise run nauth:install-examples
```
henriropp: Do not automatically install all examples in local, as they generate a lot of logs.

internal/adapter/inbound/controller/account_export.go
```diff
@@ -54,25 +54,21 @@ func NewAccountExportReconciler(k8sClient client.Client, scheme *runtime.Scheme,
...
 			return ctrl.Result{}, nil
 		}
 
 		log.Error(err, "Failed to get resource")
-		statusWrapper.setFailed(err)
-
-		if updateErr := r.Status().Update(ctx, state); updateErr != nil {
-			log.Error(updateErr, "Failed to update status", "namespace", state.Namespace, "name", state.GetName())
-			return ctrl.Result{}, updateErr
-		}
```
henriropp, line 72: Can not update status of of resource that we are not sure we managed to read.

internal/adapter/inbound/controller/account_export_status.go
```diff
@@ -90,41 +95,27 @@ func (s *accountExportStatus) setAdoptedByAccount() {
 	s.evaluateReadyCondition()
 }
 
-func (s *accountExportStatus) setFailed(err error) {
-	s.failed = err
-	s.evaluateReadyCondition()
-}
-
```
henriropp, line 97: unused

.mise-tasks/nauth/install.sh
choufraise: **suggestion**: Either remove the line entirely (if this is not temporary) or add a comment adjacent to it commented line explaining why it is just "disabled".

internal/adapter/inbound/controller/account_export.go
choufraise: **issue**: Has this been verified to be a bug or is it a hunch? We are doing exactly the same in the other controllers, hence I believe we should continue following the same pattern in this controller, and instead create an issue to investigate this, or if you are 100% sure of this, then apply it to the other controllers as well. 

In the future, I'd recommend doing such changes as a separate commit with the reason inside the commit instead of as a comment in the PR. The PR is "lost" after merge, but the commit history (blame) will be present from a devs perspective.

internal/adapter/inbound/controller/account_export_test.go
```diff
@@ -111,6 +112,104 @@ func (t *AccountExportControllerTestSuite) Test_Reconcile_ShouldFail_WhenAdopted
...
+			Name:      t.accountExportName,
+			Namespace: t.accountExportNamespace,
+		},
+		Spec: v1alpha1.AccountExportSpec{
+			AccountName: "my-account",
+			Rules:       invalidRules,
+		},
+	}
+	t.accountExportManagerMock.mockCreateClaim(t.ctx, nil, nil, fmt.Errorf("invalid rules"))
+	t.Require().NoError(k8sClient.Create(t.ctx, accountExportResource))
```
choufraise: **nit**: since this is the only place we are using `accountExportResource` we could inline it instead, making the test easier to read.

internal/adapter/inbound/controller/account_export_test.go
```diff
@@ -111,6 +112,104 @@ func (t *AccountExportControllerTestSuite) Test_Reconcile_ShouldFail_WhenAdopted
...
+	t.Require().Nil(accountExport.Status.DesiredClaim)
+
+	conditions := accountExport.Status.Conditions
+	t.Require().NotEmpty(conditions)
+
+	t.assertCondition(conditions, conditionTypeValidRules, metav1.ConditionFalse, conditionReasonInvalid)
+	t.assertCondition(conditions, conditionTypeReady, metav1.ConditionFalse, conditionReasonNotReady)
+}
+
+func (t *AccountExportControllerTestSuite) Test_Reconcile_ShouldFail_WhenExportRuleValidationFailsKeepingLastValid() {
```
choufraise, line 155: **praise**: Very nice test verifying the intention of this PR! ⭐

internal/core/account_export.go
```diff
@@ -17,8 +21,76 @@ func NewAccountExportManager() *AccountExportManager {
...
 func (a *AccountExportManager) CreateClaim(ctx context.Context, state *v1alpha1.AccountExport) (*domain.AccountExportClaim, error) {
-	// TODO: [#22] Implement AccountExportManager.CreateClaim
-	return nil, fmt.Errorf("[#22] AccountExportManager.CreateClaim not implemented")
+	exports := &jwt.Exports{}
+	for _, r := range state.Spec.Rules {
+		exports.Add(mapToJwtExport(r))
+	}
+
+	results := &jwt.ValidationResults{}
+	// NOTE: validation in jwt.Exports does not detect duplicate subjects, if the type is different.
```
choufraise: **nit**: Is this note really needed? There might be many more "bugs" in the JWT lib that we don't control, hence this note is a snapshot comment based on the _current_ jwt/v2 validation implementation. If this is fixed, then this note instantly becomes obsolete. If you want a note about this, then note that we rely 100% on the jwt/v2 validation logic.

internal/core/account_export.go
```diff
@@ -17,8 +21,76 @@ func NewAccountExportManager() *AccountExportManager {
...
+	exports.Validate(results)
+
+	warnings := results.Warnings()
+	if warnings != nil {
+		log := logf.FromContext(ctx)
+		log.Info("validation warnings for one or more exports: " + strings.Join(warnings, ", "))
```
choufraise, line 35: **suggestion**: This log will print no context, hence for larger k8s clusters with many NAuth Accounts it may be close to impossible to identify the `AccountExport` resource bound to those warnings. Suggestion is to include the namespace + name of `state`.

internal/core/account_export.go
```diff
@@ -17,8 +21,76 @@ func NewAccountExportManager() *AccountExportManager {
...
+
+	rules := make([]v1alpha1.AccountExportRule, len(state.Spec.Rules))
+	for i := range state.Spec.Rules {
+		state.Spec.Rules[i].DeepCopyInto(&rules[i])
+	}
+
+	return &domain.AccountExportClaim{Rules: rules}, nil
+}
+
+func mapToJwtExport(r v1alpha1.AccountExportRule) *jwt.Export {
```
choufraise, line 54: **suggestion**: As discussed previously, we should probably move those conversions to `domain` since we will very soon need to convert it back and forth, hence full round-trip approval tests would make sense. 

Suggestion is to add a `// TODO: [#11] Move v1alpha/jwt conversions to domain` so we don't forget it.

internal/core/account_export_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+func Test_mapToJwtExport(t *testing.T) {
+	responseThreshold := time.Duration(10)
+	tokenPos := uint(3)
+	advertise := true
+	allowTrace := true
+
+	have := v1alpha1.AccountExportRule{
+		Name:              "Name",
+		Subject:           "Subject",
```
choufraise: **suggestion**: Use an actual valid subject. Based on the `tokenPos := 3` this is not a valid rule (3 is out of bounds with the number of subject parts).

choufraise (commented):
Looking good over all, but some minor comments and suggestions.

internal/adapter/inbound/controller/account_export.go
choufraise: After offline discussion about this, I agree that we should not try to update the Status here, since we don't know the previous status (`Get` failed). The `Get` could however fail due to RBAC reasons, or any other endpoint issues, it doesn't mean that the status update will also fail. 

Since this is a new controller, it's actually not really a "change", so I'll let this pass. :) However, we should really then sync the other controllers accordingly in a separate PR imo.

.mise-tasks/nauth/install.sh
henriropp: Will remove it

internal/adapter/inbound/controller/account_export.go
henriropp: I will create a separate PR to update the other controllers

internal/core/account_export.go
henriropp: I'll remove it

internal/core/account_export.go
henriropp: I tested it and actually it does have context:

This log:
`log.Info("test message")`

Outputs this:
`2026-04-15T11:43:57Z    INFO    test message    {"controller": "accountexport", "controllerGroup": "nauth.io", "controllerKind": "AccountExport", "AccountExport": {"name":"my-account-export","namespace":"account-export-test"}, "namespace": "account-export-test", "name": "my-account-export", "reconcileID": "a155
b9d2-41d4-404d-9ff3-f3418193fca8"}`

I am guessing context is added the go context and when we get the logger:
`log := logf.FromContext(ctx)`

we get the context injected into the logs.

choufraise: Ah ok! That's nice, then we should really _stop_ adding namespace + name in all existing logs instead! 😁 👍  Carry on soldier.

choufraise (approved):

PR #222, opened 2026-04-15
chore(deps): bump Go dependencies and enforce Go version policy

choufraise:
## Summary

This PR updates Go dependencies and introduces a clear Go version policy for the repo.

## Changes

- bump Go module dependencies in go.mod / go.sum
- update the preferred Go toolchain to 1.26.2
- keep go.mod as the minimum supported Go version
- align the preferred toolchain in:
    - mise.toml
    - Dockerfile
- add hack/verify-go-version-sync.sh
- add CI validation in .github/workflows/go-version-sync.yml
- wire the version check into the Makefile

## Version policy

- go.mod defines the minimum supported Go version
- mise.toml and Dockerfile define the preferred exact toolchain version
- the preferred toolchain must stay in sync across local and Docker workflows
- the preferred toolchain must be greater than or equal to the minimum in go.mod

## Why

We had drift between the Go versions used by module metadata, local tooling, and Docker builds.

This PR fixes that by:

- keeping compatibility concerns in go.mod
- keeping the secure/default build toolchain pinned separately
- failing early in CI and make when those definitions drift

## Notes

- godebug default=go1.25 is intentionally unchanged
- this controls runtime compatibility defaults, not the selected toolchain version


go.mod
```diff
@@ -1,16 +1,16 @@
 module github.com/WirelessCar/nauth
 
-go 1.25.5
+go 1.26.2
```
eufriction: This is the minimum go version required to compile the packages. https://go.dev/doc/modules/gomod-ref#go Limiting this if not necessary reduce the ability to share the code. That might be OK but want to make you aware of the fact.

henriropp (dismissed):

go.mod
choufraise: Good catch! I'll ensure to exclude go.mod from the sync verification.

choufraise: Done. Please have a second look.

eufriction (approved):
LGTM

PR #225, opened 2026-04-15
fix: return get errors directly from reconcilers

henriropp:
## Summary
Stop routing initial `Get` failures through the status reporter in the Account, User, and NatsCluster reconcilers.

When the reconciler cannot fetch the resource and the error is not a `NotFound`, there is no safe object state to report against. Returning the raw error directly keeps controller-runtime retry behavior intact and avoids trying to write status for resources that were never successfully loaded.


choufraise (approved):
LGTM. 

A bit sad that this change didn't affect a single unit-, integration- nor e2e test though.

PR #227, opened 2026-04-16
feat(account): collect bound AccountExports as aggregated Account inputs

choufraise:
## Summary

This PR prepares Account reconciliation to consume AccountExport resources as child inputs.

It also removes the previous optimization where reconciliation work could be skipped when the Account generation and operator version had not changed. That assumption is no longer
valid once reconciliation depends on external child resources.

Issue: #11 
Related: #228 

## Changes

- move AccountExport lookup into the controller layer instead of introducing it as an outbound port
- collect bound AccountExport resources during Account reconciliation when the Account already has an account ID
- pass discovered exports into AccountResources before calling the account manager
- remove the guard based on generation == observedGeneration and operator version
- add/update tests for the reconciliation flow around bound AccountExport resources

## Important change

The reconciler can no longer assume that an unchanged Account resource means there is nothing to do.

That was true when reconciliation only depended on the Account itself. It is no longer true once AccountExport and other child CRDs become inputs to the final aggregated Account.

Because of that, this PR removes the short-circuit based on:

- metadata.generation == status.observedGeneration
- matching NAuth operator version

That change is intentional and required for correctness.

## Why

Account is moving toward an aggregated model where child CRDs contribute input to the final Account JWT.

To support that, the parent reconciler must evaluate external bound resources even when the Account spec itself has not changed.

## Notes

This PR is still preparatory:

- it does not yet implement full AccountExport adoption/status handling in Account
- it establishes the reconciliation flow needed for that next step


internal/adapter/outbound/k8s/suite_test.go
```diff
@@ -94,6 +97,15 @@ func TestMain(m *testing.M) {
 	os.Exit(code)
 }
 
+func generateNamespace() domain.Namespace {
+	value := fmt.Sprintf("%s-%d", testNamespace, time.Now().UnixNano())
+	result := domain.Namespace(value)
+	if err := result.Validate(); err != nil {
+		panic("generated invalid namespace: " + err.Error())
```
henriropp: panic?

henriropp (dismissed):
👍

internal/adapter/inbound/controller/account.go
```diff
@@ -41,17 +42,24 @@ import (
...
+	accountExportReader outbound.AccountExportReader
+	reporter            *statusReporter
 }
 
-func NewAccountReconciler(k8sClient client.Client, scheme *runtime.Scheme, manager inbound.AccountManager, recorder events.EventRecorder) *AccountReconciler {
+func NewAccountReconciler(
+	k8sClient client.Client,
+	scheme *runtime.Scheme,
+	manager inbound.AccountManager,
+	accountExportReader outbound.AccountExportReader,
```
kejne: **issue:** This introduces that the inbound adapter does direct calls to the outbound adapter. Consider if this should be part of the service logic.

internal/adapter/outbound/k8s/suite_test.go
choufraise: Yes, it's a test utility method, really should never go wrong, but it doesn't have access to the Test scope, so can't assert/require on it nor did I want to handle (test) errors that never happens.

internal/adapter/inbound/controller/account.go
choufraise: Great question! We discussed this shortly, and due to the new stricter relationships between e.g. `Account` and `AccountExport` CRDs we said that such lookup probably makes more sense closer to k8s than core. We will more or less introduce 2-way-communication between controllers via CRD status, including setting multiple `Condition`s on `AccountExport`, where one is based upon the _adoption_ from `Account.Status.Adoptions`. Of course, we also need to supply the `Account.Status.Adoptions` from the manager, which are heavily reliant on k8s concepts. 

Let's consider this PR blocked until we've sorted this out.

choufraise: Issue #228 created to untangle this mess. Will refactor this PR accordingly.

internal/core/account.go
henriropp, line 85: Isn't this accountRef validation unnecessary? the account is looked up in reconciler and passed in to manager.

choufraise: This code has nothing to do with this PR.

henriropp (approved):

Issue #228, opened 2026-04-16
Untangle the contract between inbound controllers and outbound Kubernetes adapters


choufraise:
## Summary

Refactor NAuth resource lookups so the controller layer no longer depends on outbound Kubernetes ports/adapters through controller-specific contracts.

## Problem

Today AccountReader and AccountExportReader live as outbound ports with Kubernetes implementations under internal/adapter/outbound/k8s/.

That creates a hard contract between:

- inbound controllers
- outbound Kubernetes adapters

The contract is not generic. It relies on reconciliation-specific rules and Kubernetes-specific details such as labels and lookup conventions for NAuth-
owned resources.

This is the wrong boundary. These lookups are part of controller orchestration, not reusable outbound capabilities.

## Proposed change

- remove AccountReader as an outbound port
- remove AccountExportReader as an outbound port
- move these NAuth resource lookups into the controller layer
- let controllers use the Kubernetes client directly for these queries
- keep NatsClusterReader as an outbound port

## Why NatsClusterReader stays

NatsClusterReader still represents a valid outbound abstraction.

A NatsCluster does not have to come from Kubernetes forever. It could come from another backend such as a database or external service. That makes it a
reasonable port to keep.

Account and AccountExport are different. They are NAuth-owned Kubernetes resources and their lookup rules are tightly bound to reconciliation behavior in
this controller.

## Expected outcome

- clearer separation of responsibilities
- less coupling between inbound and outbound layers
- fewer Kubernetes-specific controller contracts hidden behind ports
- easier evolution of reconciliation logic and resource lookup rules

## Acceptance criteria

- AccountReader is removed from the outbound port layer
- AccountExportReader is removed from the outbound port layer
- controller reconciliation performs these lookups directly
- NatsClusterReader remains unchanged as an outbound port
- related tests are updated
- unused adapter code is removed

PR #229, opened 2026-04-16
fix: patch AccountExport labels with metadata-only merge patch

choufraise:
Update the `AccountExport` controller to persist the bound account ID label with a metadata-only patch and continue reconciliation using the same state, instead of relying on a follow-up reconcile from metadata changes.

This keeps label updates narrow while avoiding dependence on controller event predicates for progress.

Issue: #228

### DIFF
--- a/internal/adapter/inbound/controller/account_export.go
+++ b/internal/adapter/inbound/controller/account_export.go
@@ -18,6 +18,8 @@ package controller
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 
 	"github.com/WirelessCar/nauth/api/v1alpha1"
 	"github.com/WirelessCar/nauth/internal/domain"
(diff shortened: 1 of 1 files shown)


internal/adapter/inbound/controller/account_export.go
```diff
@@ -118,3 +120,24 @@ func (r *AccountExportReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			},
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to generate patch for label: %w", err)
+	}
+	if err = r.Patch(ctx, resource, client.RawPatch(types.MergePatchType, patchData)); err != nil {
+		return fmt.Errorf("failed to patch label: %w", err)
+	}
+	// Refresh resource to ensure correct state during reconciliation
```
choufraise: **issue**: (note to self) This could be a potential future issue. Since we are refreshing the `resource`, by replacing all it's content with the latest, then any cached internal references will be detached from the resource. Due to this, the `statusWrapper` is actually working on an old version of the status.

choufraise:
Closing, will be incorporated together with #231.

PR #230, opened 2026-04-17
feat: avoid re-uploading unchanged Account JWTs to NATS

choufraise:
Add deterministic hashing of normalized NATS Account claims and persist the hash in Account status so reconciliation can detect when the effective Account JWT content is unchanged.

This is needed because the previous short-circuit based on Account generation, observedGeneration, and operator version is no longer valid once Account reconciliation depends on external child resources such as AccountExport. As reconciliation becomes more frequent, we must avoid uploading the same Account JWT to NATS over and over and unnecessarily hammering the cluster.

Issue: #11


henriropp (approved):
Nice!

PR #231, opened 2026-04-17
feat: resolve AccountExport status from account binding and adoption state

henriropp:
## Summary

  Refactor AccountExport reconciliation to compute status from a single resolution flow.

  ## Changes

  - add AccountExportManager.Resolve to evaluate claim validity, account binding, and adoption state
  - derive AccountExport conditions and desired claim from the resolved state
  - read adoption readiness from Account.status.adoptions
  - simplify controller status handling and update regression tests for the new flow

## Todo

- leave follow-up work for controller watches so Account adoption/status changes can requeue affected AccountExport resources automatically
- Manual e2e test of whole export reconciliation


internal/adapter/inbound/controller/account_export_mapper.go
```diff
@@ -0,0 +1,118 @@
...
+import (
+	"fmt"
+
+	"github.com/WirelessCar/nauth/api/v1alpha1"
+	"github.com/WirelessCar/nauth/internal/domain"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func mapResolutionToStatus(resolution *domain.AccountExportResolution, state *v1alpha1.AccountExport) (patchLabels bool, updateStatus bool) {
```
choufraise: **suggestion**: Put this method in the same file (`account_export.go`). It would keep them more tied together, and still the resulting file would not exceed 250 lines of code even.

internal/adapter/inbound/controller/account_export_mapper.go
```diff
@@ -0,0 +1,118 @@
...
+			updateConditionFalse(conditionTypeBoundToAccount, conditionReasonErrored, msg)
+		}
+
+	case domain.AccountBindingStateConflict:
+		msg := fmt.Sprintf("Account ID conflict: previously bound to %s, now found %s", resolution.BoundAccountID, resolution.AccountID)
+		updateConditionFalse(conditionTypeBoundToAccount, conditionReasonConflict, msg)
+
+	case domain.AccountBindingStateBound:
+		updateConditionTrue(conditionTypeBoundToAccount)
+	}
```
choufraise: **suggestion**: Add a `default` case where we update the condition to false with "Unknown" reason. If a new `BindingState` will be introduced, this code would still compile but we would never touch the condition. The only thing that would happen would be that we get a lint warning about not covering all cases. Since the BindingState is an enum, I see a high risk of just adding it to the manager and forgetting about this 1-1 mapping.

internal/adapter/inbound/controller/account_export_mapper.go
```diff
@@ -0,0 +1,118 @@
...
+			updateConditionFalse(conditionTypeAdoptedByAccount, conditionReasonErrored, msg)
+		} else {
+			updateConditionFalse(conditionTypeAdoptedByAccount, conditionReasonErrored, resolution.AdoptionError.Error())
+		}
+
+	case domain.AccountAdoptionStateNotAdopted:
+		updateConditionFalse(conditionTypeAdoptedByAccount, conditionReasonReconciling, resolution.AdoptionError.Error())
+
+	case domain.AccountAdoptionStateAdopted:
+		updateConditionTrue(conditionTypeAdoptedByAccount)
```
choufraise: **note**: The [comment on `BindingState` switch statement](https://github.com/WirelessCar/nauth/pull/231/changes#r3121884169) applies here as well.

internal/adapter/inbound/controller/account_export_mapper.go
```diff
@@ -0,0 +1,118 @@
...
+	if isAccountExportReady(state) {
+		updateConditionTrue(conditionTypeReady)
+	} else {
+		updateConditionFalse(conditionTypeReady, conditionReasonReconciling, "Reconciling")
+	}
+
+	return patchLabels, updateStatus
+}
+
+func newCondition(condType string, status metav1.ConditionStatus, reason string, msg string) metav1.Condition {
```
choufraise: **question**: Didn't we need the `Condition.ObservedGeneration` no longer?

internal/adapter/inbound/controller/mocks_test.go
```diff
@@ -23,10 +23,4 @@ func (a *accountReaderMock) Get(ctx context.Context, accountRef domain.Namespace
 	return args.Get(0).(*v1alpha1.Account), args.Error(1)
 }
 
-func (a *accountReaderMock) mockGet(ctx context.Context, accountRef domain.NamespacedName, result *v1alpha1.Account) *mock.Call {
-	call := a.On("Get", ctx, accountRef)
-	call.Return(result, nil)
-	return call
-}
-
 var _ outbound.AccountReader = &accountReaderMock{}
```
choufraise: **check**: I don't believe this mock is used no longer and may be fully removed, right?

internal/adapter/inbound/controller/account_export_mapper.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	// account binding
+	switch resolution.BindingState {
+	case domain.AccountBindingStateMissing:
+		if resolution.AccountID != "" {
+			state.SetLabel(v1alpha1.AccountExportLabelAccountID, resolution.AccountID)
+			patchLabels = true
+			msg := fmt.Sprintf("Binding to Account ID: %s", resolution.AccountID)
+			updateConditionFalse(conditionTypeBoundToAccount, conditionReasonReconciling, msg)
```
choufraise: **check**: On line 44 we are binding the resource to the account, why is the state set to "Reconciling" and not just "OK" directly? Wouldn't this type of intermediate state require at least one more reconcile before the resource will be OK? Maybe I'm missing some edge-case here?

internal/adapter/inbound/controller/account_export_mapper.go
```diff
@@ -0,0 +1,118 @@
...
+	case domain.AccountAdoptionStateMissing:
+		if resolution.AccountID == "" {
+			msg := "Account not found or could not be read"
+			updateConditionFalse(conditionTypeAdoptedByAccount, conditionReasonErrored, msg)
+		} else {
+			updateConditionFalse(conditionTypeAdoptedByAccount, conditionReasonErrored, resolution.AdoptionError.Error())
+		}
+
+	case domain.AccountAdoptionStateNotAdopted:
+		updateConditionFalse(conditionTypeAdoptedByAccount, conditionReasonReconciling, resolution.AdoptionError.Error())
```
choufraise: **suggestion**: To me, setting the adoption reason to "Reconciling" sounds like an assumption that the Account reconciler is currently reconciling the adoption. This may or may not be true. I believe sticking to facts like "NotAdopted" or "Unknown" would send a better message.

choufraise (commented):
Some questions. Overall, it feels like we are putting stuff in the core "just because", ending up in many duplicated code fragments and code that is hard(er) to follow. The `account_export_mapper.mapResolutionToStatus` is longer than the `AccountExportManager.Resolve`. I would have put the resource bindings and adoptions logic in the controller layer, checking each condition independently and just stuck to the actual Rule validation/mappings in core. This sub-resource is like 80% reconcile/k8s logic and 20% "core" logic, hence I'd have no problems keeping them separated as the core doesn't really care about which accounts the export is bound to or now.

internal/adapter/inbound/controller/account_export_mapper.go
henriropp: The idea was to have the mapping in a separate file, but I will move it into account_export.

internal/adapter/inbound/controller/account_export_mapper.go
henriropp: No, there was no logic related to it. It follows the Gen on the resource. I removed it. The Gen in DesiredClaim is still there and valid use case.

internal/adapter/inbound/controller/mocks_test.go
henriropp: yes. removed.

internal/adapter/inbound/controller/account_export_mapper.go
henriropp: In this specific case, the patching will require at least one more reconcile (as patching requeues the resource). 

I will go through all condition states and reasons once I have implemented the missing watches and got it working. As we discussed, there might be room for consolidating some of the condition states. But I think  need to visualize the state flow in someway first before changing...

internal/adapter/inbound/controller/account_export_mapper.go
henriropp: I will go through the conditions later (as mentioned in previous comment).

choufraise (approved):
LGTM

PR #232, opened 2026-04-17
feat: adopt account export resources into account claims

choufraise:
Adopt `AccountExport` child resources during account `CreateOrUpdate` and include the resulting adoption status in `AccountResult.Adoptions`.

This change adds export-rule-group adoption to the account claims builder, tracks adopted/invalid export resources in the account result, and covers the behavior with parameterized suite tests using approval-based assertions for the resulting YAML output.

Issue: #11


kejne (approved):
**praise:** Approvals make this easy to review 🙌

PR #236, opened 2026-04-20
feat(api): add AccountImport CRD types

choufraise:
Introduce the AccountImport API types in `api/v1alpha1` as the import-side companion to `AccountExport`.

Issue: #11


api/v1alpha1/account_import_types.go
```diff
@@ -0,0 +1,161 @@
...
+	// +optional
+	LocalSubject RenamingSubject `json:"localSubject,omitempty"`
+	// Type defines whether the import is a stream or service import.
+	// +required
+	Type ExportType `json:"type,omitempty"`
+	// +optional
+	Share *bool `json:"share,omitempty"`
+	// +optional
+	AllowTrace *bool `json:"allowTrace,omitempty"`
+}
```
henriropp, line 98: Maybe use composite struct here?
AccountImportRule

henriropp (dismissed):
Nice!

api/v1alpha1/account_import_types.go
choufraise: Fixed. Amended the commit and force-pushed it since it really made sense. Great suggestion!

henriropp (approved):

PR #237, opened 2026-04-20
build: fix broken make deploy/install targets

choufraise:
Replace stale Kubebuilder `config/`-based Make targets with working commands that match the repository's current Helm- and chart-based setup.

What changed:
- proxy `make build-installer` through a canonical Mise task
- make `install` and `uninstall` operate on chart CRDs directly
- make `deploy` and `undeploy` use Helm instead of removed `config/` kustomize paths
- update Makefile help text to reflect current kubeconfig/context behavior
- remove redundant Mise wrapper tasks that did not add project-specific value

Why:
- the old targets referenced a non-existent `config/` tree and were broken
- contributors still expect these Make targets to work
- Mise remains the canonical tool where it adds value, while Make stays usable as a compatibility layer

### DIFF
--- a/.mise-tasks/nauth/build-installer.sh
+++ b/.mise-tasks/nauth/build-installer.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+# MISE description="Render a consolidated install manifest to dist/install.yaml"
+
+set -euo pipefail
+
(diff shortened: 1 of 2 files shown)


henriropp (approved):
LGTM

PR #238, opened 2026-04-20
docs: align contributor docs with current tooling

choufraise:
Update contributor-facing documentation to reflect the repository's actual tooling and layout.

What changed:
- describe Kubebuilder as the original scaffold, not the current default workflow
- document `mise` as the canonical developer tool and `make` as a compatibility layer
- fix stale command examples (`mise nats:port-forward`, `mise run nauth:generate-docs`)
- remove the outdated `config/` layout reference from repository guidelines
- fix the stale `mise` e2e task name in `AGENTS.md`

Why:
- the previous wording implied the repo still followed the standard Kubebuilder layout and make-based workflow, which is no longer true
- contributor docs should match the commands and structure that actually exist


henriropp (approved):
LGTM

PR #240, opened 2026-04-20
refactor(manager): remove unused webhook scaffold

choufraise:
Remove unused webhook server scaffolding from the controller manager startup.

What changed:
- remove webhook server initialization from `cmd/main.go`
- remove webhook certificate flags and certificate watcher setup
- stop passing a webhook server into the controller-runtime manager
- keep metrics TLS and metrics auth configuration intact
- update the `enable-http2` flag description to refer only to the metrics server

Why:
- the repository does not register any admission webhooks
- there are no webhook manifests or chart wiring for this server
- keeping the scaffold added dead startup surface area and unnecessary flags


henriropp (approved):
LGTM

PR #241, opened 2026-04-22
feat(controller): add AccountImport reconciliation scaffolding

choufraise:
Add the first end-to-end controller path for AccountImport so the resource is wired into the operator and can start participating in reconciliation. This lays the groundwork for the remaining import-specific behavior by establishing the controller flow, connecting the runtime bindings, and covering the initial reconciliation path with tests.

Issue: #11

### DIFF
--- a/internal/adapter/inbound/controller/account_import.go
+++ b/internal/adapter/inbound/controller/account_import.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/WirelessCar/nauth/api/v1alpha1"
 	"github.com/WirelessCar/nauth/internal/domain"
@@ -86,9 +87,11 @@ func (r *AccountImportReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 	meta.SetStatusCondition(state.GetConditions(), exportAccountCondition)
 
-	if err := r.upsertLabels(ctx, state, importAccountID, exportAccountID); err != nil {
+	if labelsUpdated, err := r.upsertLabels(ctx, state, importAccountID, exportAccountID); err != nil {
 		log.Error(err, "Failed to upsert labels")
 		return ctrl.Result{}, err
+	} else if labelsUpdated {
+		return ctrl.Result{RequeueAfter: time.Millisecond}, nil
 	}
 
 	// TODO: [#11] Validate rules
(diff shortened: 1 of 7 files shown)


internal/adapter/inbound/controller/account_import.go
```diff
@@ -74,3 +126,97 @@ func (r *AccountImportReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+
+	return result, metav1.Condition{
+		Status:  metav1.ConditionTrue,
+		Reason:  conditionReasonOK,
+		Message: fmt.Sprintf("Account ID: %s", accountID),
+	}
+}
+
+func (r *AccountImportReconciler) upsertLabels(ctx context.Context, resource *v1alpha1.AccountImport, importAccountID, exportAccountID string) error {
```
henriropp: suggest: set type on importAccountID

internal/adapter/inbound/controller/account_import.go
```diff
@@ -58,9 +64,55 @@ func (r *AccountImportReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	exportAccountCondition.Type = conditionTypeBoundToExportAccount
+	if exportAccount != nil {
+		exportAccountID = exportAccount.GetLabel(v1alpha1.AccountLabelAccountID)
+	}
+	meta.SetStatusCondition(state.GetConditions(), exportAccountCondition)
+
+	if err := r.upsertLabels(ctx, state, importAccountID, exportAccountID); err != nil {
+		log.Error(err, "Failed to upsert labels")
+		return ctrl.Result{}, err
+	}
```
henriropp: I had an issue in export reconciler, where patching labels conflicted with update status in same call. Issue was resolved by requeueing after patch.

**suggest**: 		return ctrl.Result{RequeueAfter: time.Second * 2}, nil

choufraise: Separate [commit](https://github.com/WirelessCar/nauth/pull/241/commits/e9d6751d05f4ed2240ff28106792ffc4131c832d) added for this, please have a look.

choufraise: FYI, I've set the requeue to 1ms currently, we'll update this to the constant for "instant requeue" that you will add in #243 .

henriropp (approved):
LGTM

PR #242, opened 2026-04-22
refactor(core): remove Account Claims dependency towards API Spec

choufraise:
This project currently considers the `v1alpha1` API to be part of the domain model. To avoid dependencies towards more information than needed for the job we should be more explicit in the Account Claims Builder and replace the current dependency towards the full `Account.Spec` with explicit dependencies.

### DIFF
--- a/internal/core/account.go
+++ b/internal/core/account.go
@@ -150,9 +150,13 @@ func (a *AccountManager) CreateOrUpdate(ctx context.Context, resources domain.Ac
 	if err != nil {
 		return nil, fmt.Errorf("failed to get operator signing public key: %w", err)
 	}
-
-	claimsBuilder := newAccountClaimsBuilder(ctx, getDisplayName(account), account.Spec, accountPublicKey, a.accountReader)
-	claimsBuilder.addSigningKey(accountSigningPublicKey)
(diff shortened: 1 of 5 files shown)


internal/core/account_claims.go
```diff
@@ -11,175 +10,163 @@ import (
...
-				accountLimits.Conn = *spec.AccountLimits.Conn
-			}
-			if spec.AccountLimits.LeafNodeConn != nil {
-				accountLimits.LeafNodeConn = *spec.AccountLimits.LeafNodeConn
-			}
-		}
-		claim.Limits.AccountLimits = accountLimits
+	// Set default limits
+	claim.Limits.AccountLimits = jwt.AccountLimits{
+		Imports:         jwt.NoLimit,
```
henriropp, line 36: the jwt.NoLimit make much more sense :)

internal/core/account.go
```diff
@@ -488,5 +492,21 @@ func getDisplayName(account *v1alpha1.Account) string {
 	return fmt.Sprintf("%s/%s", account.GetNamespace(), account.GetName())
 }
 
+func cachedAccountIDReader(ctx context.Context, accountReader outbound.AccountReader) resolveAccountIDFn {
```
henriropp, line 495: Nice!

henriropp (approved):
LGTM

PR #243, opened 2026-04-23
feat(controller): add account and account export watches

henriropp:
## Summary

  Improve Account and AccountExport reconciliation with dependency watches, clearer status
  reporting, and deletion safeguards for accounts with exports.

  ## What changed

  - Added cross-resource watches and indexes between Account and AccountExport
  - Improved AccountExport condition reasons and status columns
  - Blocked account deletion while exports still exist
  - Updated controller and core tests, plus related cleanup

  ### Commits in this PR

  - 5f31d67 chore: cleanup logs
  - 2bccbb1 feat: block delete account if exports exist
  - 0a07c9e feat: improving condition reasons for account export
  - fc526ab feat: add TODO comments for leftover work related to AccountExport
  - a0c0fb0 feat: update AccountExport status columns to include ValidRules, Bound, and
    Adopted
  - 0e9a963 feat: use feature flag for AccountExport tests in Account reconciler
  - aff9ee8 chore: use constants for account IDs in tests
  - e7ac517 chore: update .gitignore to ignore empty codex generated files
  - 1e0117a feat: add watch for Account
  - b21515a feat: add watch for AccountExport

### DIFF
--- a/api/v1alpha1/account_export_types.go
+++ b/api/v1alpha1/account_export_types.go
@@ -31,7 +31,9 @@ const (
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
-// +kubebuilder:printcolumn:name="Message",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
+// +kubebuilder:printcolumn:name="Rules",type=string,JSONPath=`.status.conditions[?(@.type=="ValidRules")].reason`
+// +kubebuilder:printcolumn:name="Bound",type=string,JSONPath=`.status.conditions[?(@.type=="BoundToAccount")].reason`
+// +kubebuilder:printcolumn:name="Adopted",type=string,JSONPath=`.status.conditions[?(@.type=="AdoptedByAccount")].reason`
 
 // AccountExport is a component resource for exports in the accounts API.
 type AccountExport struct {
--- a/internal/adapter/inbound/controller/account.go
+++ b/internal/adapter/inbound/controller/account.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"reflect"
 
 	"github.com/WirelessCar/nauth/internal/domain"
 	"github.com/WirelessCar/nauth/internal/ports/inbound"
@@ -27,16 +28,22 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/events"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/WirelessCar/nauth/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/runtime"
-	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+const accountLabelAccountIDIndexKey = "account.labels.account-id"
 
 // AccountReconciler reconciles an Account object
 type AccountReconciler struct {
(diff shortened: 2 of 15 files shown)


api/v1alpha1/account_export_types.go
```diff
@@ -31,7 +31,9 @@ const (
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
-// +kubebuilder:printcolumn:name="Message",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].reason`
+// +kubebuilder:printcolumn:name="Rules",type=string,JSONPath=`.status.conditions[?(@.type=="ValidRules")].reason`
```
choufraise, line 34: **praise**: This is a very nice addition! 👓

internal/adapter/inbound/controller/account.go
```diff
@@ -20,38 +20,47 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/WirelessCar/nauth/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/runtime"
-	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const accountLabelAccountIDIndexKey = "account.labels.account-id"
```
choufraise: **question**: Do we really need an index for labels? The [Kubernetes docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) mentions labels and selectors as a native mechanism to relate resources and it doesn't mention any need for additional indexing.

internal/adapter/inbound/controller/account.go
```diff
@@ -117,6 +126,17 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			)
 		}
 
+		// Todo: [#11] This will block the deletion and requires manual intervention to continue (removing finalizer)
```
choufraise: **nit**: Stick with `TODO:` (all caps) since this is the standard in most languages and IDEs.

internal/adapter/inbound/controller/account.go
```diff
@@ -117,6 +126,17 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		// TODO: [#11] Investigate and understand if blocking preemptively with webhooks is a better option (not allowing change)?
+		adoptions := natsAccount.Status.Adoptions
+		if adoptions != nil && len(adoptions.Exports) > 0 {
```
choufraise, line 130: **suggestion**: We could add an `HasAdoptions()` on the `Status` struct that will be updated in one place when we add other types of adoptions like `Imports`, `SigningKeys` etc.

internal/adapter/inbound/controller/account.go
```diff
@@ -209,14 +229,18 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 func (r *AccountReconciler) collectAccountResources(ctx context.Context, account *v1alpha1.Account, accountID string) (*domain.AccountResources, error) {
 	result := domain.AccountResources{Account: *account}
-	if accountID != "" {
-		namespace := domain.Namespace(account.Namespace)
-		exports, err := r.findExportsByAccountID(ctx, namespace, accountID)
-		if err != nil {
-			return nil, err
+
+	if r.features.AccountExportEnabled {
```
choufraise, line 231: **suggestion**: Since Go is very strict regarding indentations, reverse this statement to avoid all lines within this clause being tied to the commit where we remove this experimental flag (due to indentation changes only).

```
if !r.feataures.AccountExportEnabled {
  return &result, nil
}
```

internal/adapter/inbound/controller/account.go
```diff
@@ -236,13 +260,123 @@ func (r *AccountReconciler) findExportsByAccountID(ctx context.Context, namespac
...
 }
+
+func (r *AccountReconciler) mapAccountExportToAccounts(ctx context.Context, obj client.Object) []reconcile.Request {
+	export, ok := obj.(*v1alpha1.AccountExport)
+	if !ok {
+		return nil
+	}
+
+	accountID := export.GetLabel(v1alpha1.AccountExportLabelAccountID)
+	if accountID == "" {
```
choufraise, line 286: **issue**: If the Account ID label would be (for any reason) removed, then any Account already adopting it wouldn't get triggered. 

This might also be related to why we do this `accountID != ""` check in `accountExportWatchPredicateForAccounts()`. 

Please explain why this is necessary, if it's not simply an attempt on optimization.

internal/adapter/inbound/controller/account.go
```diff
@@ -236,13 +260,123 @@ func (r *AccountReconciler) findExportsByAccountID(ctx context.Context, namespac
...
+		return nil
+	}
+
+	accounts := &v1alpha1.AccountList{}
+	if err := r.List(ctx, accounts,
+		client.InNamespace(export.Namespace),
+		client.MatchingFields{accountLabelAccountIDIndexKey: accountID},
```
choufraise: **issue**: k8s has an explicit matcher for labels, why aren't we using it? E.g:
```
client.MatchingLabels{string(v1alpha1.AccountLabelAccountID): accountID}
```

internal/adapter/inbound/controller/account_export.go
```diff
@@ -72,7 +81,9 @@ func (r *AccountExportReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	account := &v1alpha1.Account{}
 	if err := r.Get(ctx, types.NamespacedName{Namespace: state.Namespace, Name: state.Spec.AccountName}, account); err != nil {
-		log.Error(err, "Failed to get account", "accountName", state.Spec.AccountName)
+		if !apierrors.IsNotFound(err) {
+			log.Error(err, "Failed to read account", "accountName", state.Spec.AccountName)
```
choufraise, line 84: **question**: Normally when `!apierrors.IsNotFound` is `true` we fail. In this case we instead ignore the fact that we were unable to Get the Account, meaning we will continue treating the account as not ready and also _empty_ (no meta data nor labels will exist on this empty struct). Isn't this errorprone?

internal/adapter/inbound/controller/account_export.go
```diff
@@ -83,32 +94,140 @@ func (r *AccountExportReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		// Todo: #11 This would make watches simpler (not needing to be so specific)
 		if updateErr := r.Status().Update(ctx, state); updateErr != nil {
 			log.Error(updateErr, "Failed to update status")
 			return ctrl.Result{}, updateErr
 		}
 	}
 
-	log.Info("Reconciliation complete")
+	if meta.IsStatusConditionTrue(state.Status.Conditions, conditionTypeReady) {
+		log.Info("AccountExport reconciliation Ready")
```
choufraise, line 112: **nit**: This info log should probably be removed in the future since if you have e.g. X `AccountExport`s, and later a bunch of other adoptions like `AccountImport`s or `AccountSigningKey`s, then this log will be stated each time the account changes its adoption status (worst case X + Y + Z times).

internal/adapter/inbound/controller/account_export.go
```diff
@@ -83,32 +94,140 @@ func (r *AccountExportReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 		}).
+		Watches(
+			&v1alpha1.Account{},
+			handler.EnqueueRequestsFromMapFunc(r.mapAccountToAccountExports),
+			builder.WithPredicates(accountExportAccountWatchPredicate()),
+		).
 		Complete(r)
 }
 
+func exportBySpecAccountNameIndexFunc(rawObj client.Object) []string {
```
choufraise, line 141: **question**: Why don't we only watch the account that is labelled with the bound Account ID? Do we care about any other accounts?

internal/adapter/inbound/controller/account_export.go
```diff
@@ -83,32 +94,140 @@ func (r *AccountExportReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+}
+
+func accountUpdateAffectsAccountExports(oldAccount *v1alpha1.Account, newAccount *v1alpha1.Account) bool {
+	if oldAccount == nil || newAccount == nil {
+		return false
+	}
+
+	oldAccountID := oldAccount.GetLabel(v1alpha1.AccountLabelAccountID)
+	newAccountID := newAccount.GetLabel(v1alpha1.AccountLabelAccountID)
+	// this is only possible if account is deleted and recreated with new name
```
choufraise: **nit**: This is not truly true. If I'd go in and change the Account Root Seed secret, and then update the spec.label for Account ID, then I would be able to actually change the Account ID of an existing resource. Either way, I'd avoid comments like this if it's not set in stone.

internal/adapter/inbound/controller/account_export.go
```diff
@@ -83,32 +94,140 @@ func (r *AccountExportReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to patch labels: %w", err)
 		}
-		return ctrl.Result{RequeueAfter: time.Second * 2}, nil
+		// Todo: #11 Can we patch without returning, and update status in same reconcile loop?
+		return ctrl.Result{RequeueAfter: time.Millisecond * 250}, nil
```
choufraise: **suggestion**: Either set this to 1ms, which implicitly tells you that it should be as short as possible, or add a comment to _why_ it's as high as 250ms. 250 is a magic number that no one will ever dare to change without spending hours investigating it.

internal/adapter/inbound/controller/account_export_watch_test.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestAccountExportReconciler_ShouldReconcileForAccountUpdate(t *testing.T) {
```
choufraise, line 17: **issue**: First first **praise**, I really like this test! 💪 

...but, I also see a possible issue with it due to future changes. Let's face it, this test will most probably never be synced with changes done to the `Account` reconciliation logic nor CRD structure. Personally, I would rather have a couple of extra (idempotent) reconciliations that does nothing (don't touch labels, status; nothing) which would never trigger anything, than for us getting weird bugs based on this complex logic.

internal/adapter/inbound/controller/account_export_watch_test.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+	}
+
+	tests := []struct {
+		name          string
+		mutate        func(account *v1alpha1.Account)
+		expectRequeue bool
+	}{
+		{
+			// this is only possible if account is deleted and recreated with new name
```
choufraise: **nit**: Please remove this comment, it may not be always true.

internal/adapter/inbound/controller/account_watch_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestAccountReconciler_ShouldReconcileForAccountExportUpdate(t *testing.T) {
```
choufraise, line 17: **issue**: Same concern as for the `TestAccountExportReconciler_ShouldReconcileForAccountUpdate` in `account_export_watch_test.go`. Those two tests will have always have to be in sync with each other and each respective reconcile logic, otherwise we _will_ get very strange bugs.

choufraise (commented):
Overall it looks good and well thought-through, but I have some major concerns about the watches that I'd like to settle before approving. See comments.

internal/adapter/inbound/controller/account.go
henriropp: Ok. I will remove this index and use the `client.MatchingLabels{}` for lookup instead.

internal/adapter/inbound/controller/account.go
henriropp: I was considering that, but decided to go for exports in this change. We can refactor it when we introduce the next type of adoption.

internal/adapter/inbound/controller/account.go
henriropp: No worries, I will make sure to cleanup the experimental flag (already started on a different branch).

internal/adapter/inbound/controller/account_export.go
henriropp: I agree it is not optimal. Reason it continues is that the check is in `manager.Resolve()`. I will try to lift the check up to the controller.

internal/adapter/inbound/controller/account_export.go
henriropp: I changed the log here to only print when it is successfully (ready) reconciled, before it was printing every reconciliation loop. We can remove it completely when we have the feature completed with imports. I'll add a todo for that purpose.

internal/adapter/inbound/controller/account_export.go
henriropp: I removed it

internal/adapter/inbound/controller/account_export.go
henriropp: I lowered it from the 2s, but when I asked AI if I could set 1ms, it reaced with: 

> The delay is probably not for the patch call itself, but to give the controller cache time to observe the label update before the next reconcile. With 1ms, the next reconcile may still read a stale AccountExport from cache and try to patch again, causing noisy extra loops.

So I thought 250ms might be enough. 
I will add a constant for the value, so we get rid of the magic number.

internal/adapter/inbound/controller/account.go
henriropp: When creating an export without a label for account, we do not need to trigger the account that we do not know of.

when changing the account label for an an export, both the old and new export resource will be reconciled. The one with the label set will trigger account. the empty label one will not.

internal/adapter/inbound/controller/account_export.go
henriropp: The export references the account by name in the spec. It feels more correct to watch by what is referenced with.

According to AI, having a watch on AccountID, would miss the scenario where an account is created after an export.

internal/adapter/inbound/controller/account_export_watch_test.go
henriropp: I will add some kuttl tests also that will capture the interaction between the two resources, in a later commit

internal/adapter/inbound/controller/account_export.go
choufraise: If an `Account` is created after an `AccountExport`, the Account will get reconciled due to itself being created, hence it will lookup the `AccountExport` during that creation reconciliation loop. 🤔  However, let's keep it as-is, I guess we'll figure stuff like this out better when we have it running for real.

internal/adapter/inbound/controller/account_watch_test.go
henriropp: I will add some kuttl tests also that will capture the interaction between the two resources, in a later commit

internal/adapter/inbound/controller/account_export.go
choufraise: Ah, ok, makes sense. 👍

choufraise (dismissed):
LGTM as a first test. Agreed that we will refactor some of the reconcile logic later on.

choufraise (approved):

PR #244, opened 2026-04-24
feat!: stop default-enabling JetStream on accounts

choufraise:
## Summary

This PR introduces a breaking business-logic change to Account JetStream behavior in NAuth.

NAuth no longer treats JetStream as implicitly enabled by default on Accounts. Instead, Account reconciliation now follows the real behavior of `nats-io/jwt`: JetStream remains disabled unless account claims explicitly allow it.

To make that behavior explicit and safe, this PR adds `spec.jetStreamEnabled` and updates claim building, claim conversion, and end-to-end coverage around JetStream-enabled Accounts.

## Why

See `nats-io/jwt#249`.

During investigation together with Synadia, we confirmed that the default JetStream limits returned by `jwt.NewAccountClaims(...)` do **not** mean "JetStream enabled with unlimited limits". In practice, JetStream stays disabled unless `memStorage` and/or `diskStorage` is set to a non-zero value.

That made the previous NAuth behavior misleading: NAuth projected a more permissive default than what the JWT library and NATS server actually enforce.

Since the upstream behavior is legacy and cannot be changed safely, NAuth needs to align with it instead of masking it.

## Main Change

This PR makes JetStream enablement explicit at the NAuth Account level.

### New behavior

- `spec.jetStreamEnabled: true`
  NAuth explicitly enables JetStream by seeding unlimited `memStorage` and `diskStorage` (`-1`), plus unlimited core JetStream capacity defaults unless overridden.

- `spec.jetStreamEnabled: false`
  NAuth preserves the default JetStream limits from `jwt.NewAccountClaims(...)`, which means JetStream remains disabled.

- `spec.jetStreamEnabled` omitted
  NAuth does not force the result. Effective JetStream behavior is derived from the supplied JetStream limits using the same semantics as `nats-io/jwt`.

### Validation

This PR also rejects ambiguous configurations where the requested boolean intent conflicts with the effective JWT limits, for example:

- JetStream requested as enabled while storage limits still resolve to disabled JetStream
- JetStream requested as disabled while supplied storage limits would implicitly enable JetStream

## Other Changes In This Branch

In addition to the main breaking change, this branch also includes the earlier claim-default cleanup work:

- stop materializing default JWT limit values unnecessarily in NAuth claim conversion
- align generated / converted account claims more closely with `nats-io/jwt` defaults
- improve approval coverage around default and omitted claim fields
- strengthen the JetStream KUTTL scenario so it proves stream creation actually succeeds before testing deletion behavior

Those earlier commits exposed the mismatch between NAuth's previous assumptions and the real upstream JetStream behavior, which led to the explicit `jetStreamEnabled` model introduced here.

## Breaking Change

This is a **breaking business-logic change**, not a CRD compatibility break.

Existing manifests still deserialize, but the effective default behavior changes:

- JetStream is no longer treated as implicitly enabled for Accounts
- users who relied on the previous behavior must now set `spec.jetStreamEnabled: true` or explicitly provide JetStream storage limits that enable JetStream

## Reference

- `nats-io/jwt#249`


eufriction (dismissed):
LGTM

internal/core/approvals/account_claims_test.Test_AccountClaims.default-limits.output.nauth.approved.yaml
```diff
@@ -1,22 +1,5 @@
...
-  diskStorage: -1
-  maxAckPending: -1
-  memMaxStreamBytes: -1
-  memStorage: -1
-  streams: -1
-natsLimits:
-  data: -1
-  payload: -1
-  subs: -1
+jetStreamEnabled: false
```
choufraise, line 2: **note**: This line proves the breaking change. 

Please leave this note unresolved (blocking merge) until at least 2 other maintainers have approved this.

kejne (dismissed):
LGTM! A good addition - good to add this early. 👌

kejne (commented):
Whenever there is a breaking change, these must be introduced so that users can upgrade to one version and do the necessary updates to their CR:s until they upgrade to start using the breaking changes.

In this case, the JetStreamEnabled would need to be a no-op in first release (or CRD only update) followed by a release with the breaking change.

choufraise:
An alternative PR (to allow simpler migration before breaking change is actually introduced) is found here: #246 . Only one of those should be merged.

choufraise:
Created a ticket for this #245 .

choufraise:
PR super-seeded with #246 . Closing.

internal/core/approvals/account_claims_test.Test_AccountClaims.default-limits.output.nauth.approved.yaml
choufraise: PR closed.

PR #250, opened 2026-04-29
refactor(core): model account adoptions in nauth domain

choufraise:
## Summary

Moves Account adoption handling into the NAuth domain model and keeps Kubernetes status formatting at the controller boundary.

Issue: #11

## What changed

- Added domain structures for Account adoptions, adoption results, export groups, exports, response types, and service latency
- Changed core account reconciliation to produce domain adoption results
- Converts AccountExport desired claims into domain export groups before claim building
- Converts domain adoption results back into Account CRD status adoptions
- Updated controller/core tests and approval snapshots

PR #252, opened 2026-04-29
refactor(export): move account export resolution into the controller

henriropp:
## Description

This refactor narrows the core account export manager to rule validation and JWT export mapping, while moving account binding, adoption tracking, and desired-claim/status handling into the AccountExport controller.

It also introduces domain-level export rule types and updates the test suite to cover the new controller-owned reconciliation flow and the slimmer core validation behavior.

### DIFF
--- a/internal/adapter/inbound/controller/account_export.go
+++ b/internal/adapter/inbound/controller/account_export.go
@@ -23,7 +23,7 @@ import (
 	"reflect"
 
 	"github.com/WirelessCar/nauth/api/v1alpha1"
-	"github.com/WirelessCar/nauth/internal/domain"
+	"github.com/WirelessCar/nauth/internal/domain/nauth"
 	"github.com/WirelessCar/nauth/internal/ports/inbound"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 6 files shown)


internal/domain/nauth/export.go
```diff
@@ -0,0 +1,34 @@
...
+	// +optional
+	Name string
+	// +required
+	Subject Subject
+	// +required
+	Type ExportType
+	// +optional
+	ResponseType ResponseType
+	// +optional
+	ResponseThreshold *time.Duration
```
choufraise: **nit**: The representation of `time.Duration` is just an `int64`, meaning default will be zero and the pointer is useless. JWT lib doesn't use pointer here for instance.

internal/adapter/inbound/controller/account_export.go
```diff
@@ -78,40 +78,157 @@ func (r *AccountExportReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	for _, rule := range state.Spec.Rules {
+		rules = append(rules, mapToExportRule(rule))
+	}
+
+	err := r.manager.ValidateRules(ctx, rules)
+	if err != nil {
+		updateConditionFalse(conditionTypeValidRules, conditionReasonInvalid, err.Error())
+	} else {
+		updateConditionTrue(conditionTypeValidRules)
+		(&state.Status).DesiredClaim = &v1alpha1.AccountExportClaim{
```
choufraise: **check**: Is this `(&state.Status)` really necessary?

internal/adapter/inbound/controller/account_export.go
```diff
@@ -78,40 +78,157 @@ func (r *AccountExportReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			updateConditionFalse(conditionTypeBoundToAccount, conditionReasonFailed, msg)
+		} else {
+			// account not found
+			msg := "Account not found"
+			updateConditionFalse(conditionTypeBoundToAccount, conditionReasonNotFound, msg)
+		}
+	} else {
+		// account binding
+		accountID := account.GetLabel(v1alpha1.AccountLabelAccountID)
+		state.Status.AccountID = accountID
```
choufraise, line 121: **check**: What happened to the `updateStatus bool`? Right now the status will always be updated and might implicitly always trigger other resources to reconcile, is this the intentional?

internal/adapter/inbound/controller/account_export.go
```diff
@@ -78,40 +78,157 @@ func (r *AccountExportReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		if adoption == nil {
+			// adoption not found
+			updateConditionFalse(conditionTypeAdoptedByAccount, conditionReasonAdopting, "waiting for account to adopt export")
+		} else {
+			adoptionGen := adoption.Status.DesiredClaimObservedGeneration
+			sameGeneration := adoptionGen != nil && *adoptionGen == state.Generation
+			if adoption.Status.Status == metav1.ConditionTrue && sameGeneration {
+				updateConditionTrue(conditionTypeAdoptedByAccount)
+			} else {
+				if !sameGeneration {
```
choufraise: **suggestion**: `else if`

internal/core/account_export.go
```diff
@@ -131,36 +51,35 @@ func mapToJwtExport(r v1alpha1.AccountExportRule) *jwt.Export {
...
 }
 
-func findAdoption(state *v1alpha1.AccountExport, account *v1alpha1.Account) (*v1alpha1.AccountAdoption, error) {
-	if found, adoption := findAdoptionByUID(account, state.UID); found {
-		return &adoption, nil
-	}
-
-	return nil, fmt.Errorf("account export is not yet processed by account")
-}
+func validateRules(ctx context.Context, exports *jwt.Exports) error {
```
choufraise, line 63: **note**: I put validations like this in the `account_claims.go` to ensure we use the same validation everywhere. This is easy enough to refactor later though, so let's keep it like this.

choufraise (dismissed):
LGTM. Some conflicts with my pending PR #250 , but we are pretty aligned so it shouldn't be that hard to merge those conflicts.

internal/adapter/inbound/controller/account_export.go
henriropp: nope, weird, maybe an inlining op... Removed it.

internal/domain/nauth/export.go
henriropp: updated. I wanted to follow the format of the export in jwt, but missed that one.

internal/adapter/inbound/controller/account_export.go
henriropp: updateStatus was always true before (not implemented). 
I checked, if the status is same, no updates will happen (on the resource) , and thus no triggering.

internal/adapter/inbound/controller/account_export.go
henriropp: joined

choufraise (approved):

PR #253, opened 2026-04-30
fix(core): deduplicate grouped imports and exports

choufraise:
## Summary

Allows Account claim generation to merge identical imports and exports from multiple groups without failing validation.

Refs #11.

## What changed

- Deduplicates identical imports and exports during grouped claim merging
- Keeps standalone AccountImport and AccountExport validation strict for duplicates in the same input
- Adds regression coverage for duplicate validation vs grouped merge behavior
- Cleans up NAuth domain model placement and removes unused export-resolution code
- Fixes the JetStream deletion E2E step so failures print the NATS response and retry briefly


internal/core/account_claims.go
```diff
@@ -285,74 +285,62 @@ func convertNatsAccountClaims(claims *jwt.AccountClaims) nauth.AccountClaims {
...
+	})
 }
 
-func mergeJWTImports(importAccountID string, existing jwt.Imports, extras jwt.Imports) (jwt.Imports, error) {
-	tmpResult := existing
-	appendIfMissing := func(haystack jwt.Imports, needle jwt.Import) jwt.Imports {
-		for _, e := range haystack {
-			if e != nil && reflect.DeepEqual(*e, needle) {
-				return haystack
+func mergeJWTItems[T jwt.Import | jwt.Export](existing []*T, additions []*T, mergeDuplicates bool, valFn func([]*T, *jwt.ValidationResults)) ([]*T, error) {
```
henriropp: I am thinking that this function might be doing too much. It merges and validates. Maybe validation could be its own function.

choufraise: Great comment! ⭐ I've refactored this completely, it ended up much cleaner imo.

henriropp: Yes, looks nice now :smiley:

henriropp (approved):

PR #255, opened 2026-05-04
refactor(core): move NatsCluster API handling to adapters

choufraise:
Break the core cluster flow away from `v1alpha1.NatsCluster` and `v1alpha1.NatsClusterRef` by introducing domain-level `nauth.ClusterRef` and `nauth.ClusterTarget` models.

Move Kubernetes-specific NatsCluster resolution into the outbound k8s adapter, so core validation only receives a resolved cluster target and account flows only pass domain cluster references. This keeps API conversion at the boundary and removes direct NatsCluster resource handling from core.

Issue: #11

### DIFF
--- a/internal/adapter/outbound/k8s/cluster.go
+++ b/internal/adapter/outbound/k8s/cluster.go
@@ -6,32 +6,164 @@ import (
 
 	"github.com/WirelessCar/nauth/api/v1alpha1"
 	"github.com/WirelessCar/nauth/internal/domain"
+	"github.com/WirelessCar/nauth/internal/domain/nauth"
 	"github.com/WirelessCar/nauth/internal/ports/outbound"
+	"github.com/nats-io/nkeys"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-type NatsClusterClient struct {
-	reader client.Reader
+type ClusterClient struct {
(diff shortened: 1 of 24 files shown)


internal/adapter/outbound/k8s/cluster.go
```diff
@@ -6,32 +6,169 @@ import (
...
+	if err = c.k8sReader.Get(ctx, client.ObjectKey{Namespace: namespacedNameRef.Namespace, Name: namespacedNameRef.Name}, cluster); err != nil {
+		return nil, fmt.Errorf("failed getting NatsCluster resource %s: %w", namespacedNameRef.String(), err)
+	}
+
+	target, err := c.ResolveClusterTarget(ctx, cluster)
+	if err != nil {
+		return nil, err
+	}
+
+	return target, nil
```
henriropp: Should be same as: `return c.ResolveClusterTarget(ctx, cluster)`

henriropp (dismissed):
LGTM

internal/adapter/outbound/k8s/cluster.go
choufraise: Fixed and force-pushed.

henriropp (approved):

PR #259, opened 2026-05-05
feat: adopt account imports in account reconciliation

henriropp:
This PR is starting to get big, so I thought better to PR and merge...
Tested manually in local environment. Seems to reconcile imports and react to changes.

### Description

This PR adds `AccountImport` adoption to the account reconciliation flow so accounts now collect import resources, report import adoption status, and trigger reconciles when related imports or accounts change.

It also **removes the temporary import/export feature flags** and promotes both controllers to the default startup path. The change updates the `Account` and `AccountImport` APIs/CRDs, refreshes chart manifests, and adjusts controller tests and import/export scenarios to cover the new behavior.

**Replaced deprecated scheme builder** in v1alpha1 (due to linter error).

## Leftovers (coming in separate PRs)

* More unit testing needed
* KUTTL tests
* More refactoring / cleanup


api/v1alpha1/groupversion_info.go
```diff
@@ -14,23 +14,41 @@ See the License for the specific language governing permissions and
...
 
 	// SchemeBuilder is used to add go types to the GroupVersionKind scheme.
-	SchemeBuilder = &scheme.Builder{GroupVersion: GroupVersion}
+	SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
 
 	// AddToScheme adds the types in this group-version to the given scheme.
 	AddToScheme = SchemeBuilder.AddToScheme
 )
+
+func addKnownTypes(scheme *runtime.Scheme) error {
```
choufraise, line 39: **question**: What's the benefit of registering the schemas like this instead of the `init` functions as before?

internal/adapter/inbound/controller/account.go
```diff
@@ -310,13 +367,17 @@ func (r *AccountReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		&v1alpha1.AccountExport{},
+		handler.EnqueueRequestsFromMapFunc(r.mapAccountExportToAccounts),
+		builder.WithPredicates(accountExportWatchPredicateForAccounts()),
+	)
+
+	controllerBuilder = controllerBuilder.Watches(
+		&v1alpha1.AccountImport{},
+		handler.EnqueueRequestsFromMapFunc(r.mapAccountImportToAccounts),
+		builder.WithPredicates(accountImportWatchPredicateForAccounts()),
+	)
```
choufraise: **suggestion**: Inline those instead to make this easier to read as it now looks like you are (accidentally) resetting the builder all the time. Example:
```
return ctrl.NewControllerManagedBy(mgr).
    For(...).
    WithOptions(...).
    Watches(...).
    Watches(...).
    Complete(r)
```

internal/adapter/inbound/controller/account.go
```diff
@@ -407,8 +468,94 @@ type accountExportDesiredClaimState struct {
...
+			export, ok := e.Object.(*v1alpha1.AccountImport)
+			return ok && export.GetLabel(v1alpha1.AccountImportLabelAccountID) != ""
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			oldImport, oldOK := e.ObjectOld.(*v1alpha1.AccountImport)
+			newImport, newOK := e.ObjectNew.(*v1alpha1.AccountImport)
+			return oldOK && newOK && accountImportUpdateAffectsAccounts(oldImport, newImport)
+		},
+		GenericFunc: func(event.GenericEvent) bool {
+			return false
```
choufraise, line 515: **suggestion**: When are those `GenericEvent`s triggered? Could we add a comment to why we always return `false`?

internal/adapter/inbound/controller/account_import.go
```diff
@@ -159,16 +182,119 @@ func (r *AccountImportReconciler) validateImports(importAccountID string, export
...
+			Status:                         adoption.Status.Status,
+			Reason:                         adoption.Status.Reason,
+			Message:                        adoption.Status.Message,
+			DesiredClaimObservedGeneration: adoption.Status.DesiredClaimObservedGeneration,
+		}
+	}
+	return adoptions
+}
+
+type accountImportAdoptionState struct {
```
choufraise: **check**: This struct doesn't feel bound to "AccountImport". Couldn't it be a generic `AdoptionState` struct used by exports (and then Signing Keys etc) as well?

internal/adapter/inbound/controller/adoption.go
```diff
@@ -0,0 +1,34 @@
+package controller
+
+import (
+	"github.com/WirelessCar/nauth/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func findExportAdoptionByUID(adoptions *v1alpha1.AccountAdoptions, uid types.UID) *v1alpha1.AccountAdoption {
```
choufraise: **suggestion**: Use `findAdoptionByUID(...)` directly instead. The `adoptions != nil` check here makes it harder to read (from the usage perspective) than if this check was done there.

choufraise (dismissed):
LGTM

api/v1alpha1/groupversion_info.go
choufraise: Got this lint error in my other PR, resolving this. 😄 

```
Error: api/v1alpha1/groupversion_info.go:32:19: SA1019: scheme.Builder is deprecated: This helper is only useful in api packages, but api packages should be easy to import and hence have minimal dependencies. Typically, these dependencies include only the standard library, k8s.io/apimachinery and other api packages. (staticcheck)```

internal/adapter/inbound/controller/account.go
henriropp: sure. leftovers from removing the feature flags.

internal/adapter/inbound/controller/account.go
henriropp: I'll look it up

henriropp: It is all other events that are not create, update, or delete. something manual is mentioned.

internal/adapter/inbound/controller/account_import.go
henriropp: yes, I reduced it to one adoptionState. I will do additional refactoring later.

choufraise (approved):

PR #261, opened 2026-05-05
refactor(domain): use account ID reader with typed errors

choufraise:
Replace core Account lookups with AccountIDReader so core code only depends on the Account ID it needs instead of reading `v1alpha1.Account` directly.

Introduce typed domain errors for account/configmap lookup failures and remove the k8s adapter-specific errors. Update the k8s outbound adapter, mocks, and tests to use the new domain errors with errors.Is support.

### DIFF
--- a/internal/adapter/inbound/controller/user_test.go
+++ b/internal/adapter/inbound/controller/user_test.go
@@ -7,7 +7,7 @@ import (
 	"testing"
 
 	"github.com/WirelessCar/nauth/api/v1alpha1"
-	"github.com/WirelessCar/nauth/internal/adapter/outbound/k8s"
+	"github.com/WirelessCar/nauth/internal/domain"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/suite"
 	k8err "k8s.io/apimachinery/pkg/api/errors"
@@ -95,14 +95,14 @@ func (t *UserControllerTestSuite) Test_Reconcile_ShouldSucceed_WhenCreatingOrUpd
 
 func (t *UserControllerTestSuite) Test_Reconcile_ShouldFail_WhenCreateOrUpdateFailsBecauseNoAccountExists() {
 	// Given
-	t.userManagerMock.On("CreateOrUpdate", mock.Anything, mock.Anything).Return(k8s.ErrNoAccountFound).Once()
+	errAccountNotFound := domain.ErrAccountNotFound()
+	t.userManagerMock.On("CreateOrUpdate", mock.Anything, mock.Anything).Return(errAccountNotFound).Once()
 
 	// When
 	_, err := t.unitUnderTest.Reconcile(t.ctx, reconcile.Request{NamespacedName: t.userNamespacedName})
 
 	// Then
-	t.Error(err)
-	t.Equal(k8s.ErrNoAccountFound, err)
+	t.ErrorIs(err, errAccountNotFound)
 
 	user := &v1alpha1.User{}
 	err = k8sClient.Get(t.ctx, t.userNamespacedName, user)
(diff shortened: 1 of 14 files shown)


internal/adapter/inbound/controller/user_test.go
```diff
@@ -95,14 +95,15 @@ func (t *UserControllerTestSuite) Test_Reconcile_ShouldSucceed_WhenCreatingOrUpd
 
 func (t *UserControllerTestSuite) Test_Reconcile_ShouldFail_WhenCreateOrUpdateFailsBecauseNoAccountExists() {
 	// Given
-	t.userManagerMock.On("CreateOrUpdate", mock.Anything, mock.Anything).Return(k8s.ErrNoAccountFound).Once()
+	rootCause := errors.New("account not found")
+	t.userManagerMock.On("CreateOrUpdate", mock.Anything, mock.Anything).Return(rootCause).Once()
```
choufraise: **note**: The `k8s.ErrNoAccountFound` was only used in this test and has no impact what so ever on the actual code, hence replacing it with a random error.

choufraise: Actually, I'll replace it with the explicit domain error instead, brb.

cmoscofian (dismissed):

internal/domain/errors.go
henriropp, line 1: I am not so sure about new struct for domain errors. What problem is it solving?

I see ErrConfigMapNotFound, ErrAccountNotFound and ErrAccountNotReady are basically static, using the const ErrCode...

This could typically be done easier with:
```
var (
    ErrConfigMapNotFound = errors.New("ConfigMapNotFound")
    ErrAccountNotFound = errors.New("AccountNotFound")
    ...
)
```

It should still be possible to add dynamic messages, by wrapping it:
`return fmt.Errorf("accountId %q: %w", id, ErrAccountNotFound)`

and still use errors.Is:
```
if errors.Is(err, ErrAccountNotFound) {
    // handle not found
}
```

We can huddle about it tomorrow.. I am about to sign off..

choufraise: The idea was to force interfaced methods to return an `NAuthError` instead of `error`. Apparently that was not the a best practice anyway, so I removed that explicit requirement on the port interfaces. I'm still having issues with exporting `var`s as any code in any package could easily change those variables at any time which could become a vulnerability. I'll try to simplify this in a separate (V2) PR, then let's discuss tomorrow.

choufraise: I really didn't want to use `var` for those errors (due to the possible vulnerability mentioned above), but I simplified it a bit together with Codex in a separate commit. Please review.

choufraise: Found a very nice pattern imho, which kills the vulnerability of using exported vars. Please have a look.

internal/domain/errors.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func BadRequest(cause error) error {
+	return WrapError(ErrBadRequest, cause)
+}
+
+func UnknownError(cause error) error {
+	return WrapError(ErrUnknownError, cause)
+}
+
```
henriropp: Would this work, instead of the specific convenience functions?

```
func (e Error) WithCause(cause error) error {
	return WrapError(e, cause)
}

func (e Error) ToError() error {
	return e
}
```
The ToError()  might not be needed...

internal/domain/errors.go
henriropp: Nice! Much better 👍

henriropp (dismissed):
Great! This error handling is really nice. I suggested something that might improve it even more, if it works :) 
Good work!

internal/domain/errors.go
choufraise: Oooh, the builder pattern. I like it, will absolutely test it out!

henriropp: `err := domain.ErrAccountNotFound.WithCause(fmt.Errorf("account %q not found", ref))`

henriropp (approved):

PR #264, opened 2026-05-06
test(e2e): add KUTTL coverage for AccountExport workflows

henriropp:
## Summary

Add end-to-end KUTTL coverage for `AccountExport` behavior so export reconciliation is exercised across the main happy path and key failure/recovery scenarios.

## What changed

- add a new `account-export` KUTTL scenario covering:
  - account creation
  - export creation
  - export adoption by the account
  - export updates propagating into the desired claim
- add an `account-export-conflict` scenario covering competing exports for the same account and asserting conflict handling/adoption state
- add an `account-export-invalid-update` scenario covering invalid rule updates and asserting that the previously observed desired claim is preserved while reconciliation remains non-ready
- add an `account-export-missing-account` scenario covering export creation before the target account exists, then verifying recovery once the account is created
- add `kuttl-fast.yaml` to make it easier to iterate on KUTTL suites against an already-running local Kind cluster

## Why

`AccountExport` flows had limited e2e coverage. These tests lock in expected behavior for reconciliation, adoption, conflict handling, recovery, and teardown, reducing the risk of regressions in controller behavior.

## How to validate

```bash
make test-e2e

### DIFF
--- a/test/e2e/account-export/02-assert-account.yaml
+++ b/test/e2e/account-export/02-assert-account.yaml
@@ -1,0 +1,18 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 20
+resourceRefs:
+  - apiVersion: nauth.io/v1alpha1
+    kind: Account
+    name: my-account
+    ref: my_account
+  - apiVersion: nauth.io/v1alpha1
+    kind: AccountExport
+    name: my-account-export
+    ref: my_account_export
+assertAll:
+  - celExpr: my_account.status.conditions.exists(c, c.type == "Ready" && c.status == "True" && c.reason == "Reconciled")
+  - celExpr: has(my_account.status.adoptions) && has(my_account.status.adoptions.exports) && my_account.status.adoptions.exports.size() == 1
+  - celExpr: my_account.status.adoptions.exports[0].uid == my_account_export.metadata.uid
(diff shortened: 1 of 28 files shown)

### CODE REVIEW

**Review by choufraise — COMMENTED** [review]
Let's see if https://github.com/WirelessCar/nauth/pull/268 solves the 90-* issue before approving this.

**Review by choufraise — APPROVED** [review]
LGTM 🎖️  Nice!

File: test/e2e/account-export/02-assert-account.yaml [hunk 1]
```diff
@@ -0,0 +1,18 @@
+apiVersion: kuttl.dev/v1beta1
```
choufraise, line 1: **question**: Are you sure that both `02-assert-account.yaml` and `02-assert-export.yaml` are picked up by kuttl? 🤔  I _think_ I've seen issues before where I thought it picked up more than one file, but it didn't. Maybe we can verify this?

henriropp: Yes, I am sure. But I will verify it even though... :laughing:

choufraise: Sorry, you don't need to verify if you are sure. I remembered incorrectly. My mind was referring to multiple `TestStep`s in one file, then kuttl only looks at the first one. Feel free to resolve this.

henriropp: Could not find it in the docs, but I remember I looked in the code a while ago.

https://github.com/kudobuilder/kuttl/blob/main/internal/file/name.go#L50

PR #266, opened 2026-05-06
refactor(account): move Account CreateOrUpdate to domain request model

choufraise:
## Summary

Moves Account CreateOrUpdate onto the `nauth.AccountRequest` domain model so the core no longer depends on API-level `v1alpha1.Account` resources.

## What changed

- Changed `AccountManager.CreateOrUpdate` to accept `nauth.AccountRequest`.
- Removed the temporary core-side Account converter and `AccountResources` wrapper.
- Moved Account CRD to domain conversion into the account controller layer.
- Converts inline Account imports and exports into required domain import/export groups.
- Resolves inline import account IDs through the injected Kubernetes account reader.
- Keeps external `AccountExport` and `AccountImport` resources as adoption groups.
- Updated controller/core tests and approval snapshots for the domain request contract.

## Why

This keeps the core account manager working on domain types only and leaves API/CRD conversion responsibilities in the controller layer.

Related: #11


kejne (approved):
LGTM!

PR #267, opened 2026-05-06
fix(account): strict enum conversions

### LINKED ISSUE #265: API/NAuth/JWT enum conversions should handle defaults in one place alone
**Describe the bug**
Today NAuth is converting enums between `v1alpha1`, `nauth` (domain) and `jwt` (nats-io) values, and back. The conversion functions are not streamlined and some of them default to a (possibly invalid) value when an unknown or "" value is supplied as parameter. Secondly, in some places the conversion is also based on pure string value conversion, e.g. like [this](https://github.com/WirelessCar/nauth/blob/044129e5f644b0a9a6336604be64b77c3c21fd6c/internal/adapter/inbound/controller/account_converter.go#L260) `v1alpha1.ResponseType(exp.ResponseType),` which only works if the casing matches, which it doesn't.  

Related enums: 
* ResponseType
* ExportType

choufraise:
## Summary

Adds explicit conversion handling for Account export and response type enums so unsupported values fail fast instead of being silently mapped to empty or unknown values.

## What changed

- Added stricter enum conversion for API, domain, and NATS JWT export/response types.
- Propagates conversion errors through Account controller and core claim flows.
- Updates AccountImport rule validation to fail when rule enum conversion fails.
- Makes `status.claims` optional via pointer semantics.
- Regenerates deepcopy output for the API status change.
- Adds test coverage for valid, empty, unknown, and invalid enum conversions.

Closes #265


henriropp (approved):
LGTM

PR #268, opened 2026-05-06
fix(account): block account deletion while imports or exports are still bound

henriropp:
# Summary

This PR fixes account deletion handling so an `Account` cannot be removed while dependent `AccountImport` or AccountExport` bindings still exist. Previously the delete flow only relied on adoption status, which did not update once in the deletion logic flow.

* The reconciliation logic now lists bound imports and exports directly during account deletion and returns a clear error when bindings are still present. 

* It also updates AccountImport RBAC to allow the controller to patch status and labels as part of reconciliation.

### DIFF
--- a/internal/adapter/inbound/controller/account.go
+++ b/internal/adapter/inbound/controller/account.go
@@ -123,7 +123,6 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			log.Info("Failed to list users", "name", natsAccount.Name, "error", err)
 			return ctrl.Result{}, err
 		}
-
 		if len(userList.Items) > 0 {
 			return r.reporter.error(
 				ctx,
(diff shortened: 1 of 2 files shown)


internal/adapter/inbound/controller/account.go
```diff
@@ -132,14 +131,29 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+				fmt.Errorf("cannot delete an account with bound exports, found %d bindings", len(exports.Items)),
+			)
+		}
+
+		imports, err := r.findImportsByAccountID(ctx, domain.Namespace(natsAccount.Namespace), accountID)
+		if err != nil {
+			log.Error(err, "Failed to list imports for account during deletion", "name", natsAccount.Name)
+			return ctrl.Result{}, err
+		}
+		if len(imports.Items) > 0 {
```
choufraise: **suggestion**: Extract this entire block (including userLIst) into a separate method for collecting bound/dependant upon sub-resources.

choufraise (dismissed):

internal/adapter/inbound/controller/account.go
henriropp: done

internal/adapter/inbound/controller/account.go
```diff
@@ -517,6 +499,43 @@ func (r *AccountReconciler) mapAccountImportToAccounts(ctx context.Context, obj
...
+
+	// Check for bound users
+	userList := &v1alpha1.UserList{}
+	err := r.List(ctx, userList, client.MatchingLabels{string(v1alpha1.UserLabelAccountID): string(accountID)}, client.InNamespace(namespace))
+	if err != nil {
+		log.Error(err, "Failed to list users bound to account")
+		return domain.ErrUnknownError.WithCause(err)
+	}
+	if len(userList.Items) > 0 {
+		return fmt.Errorf("cannot delete an account with associated users, found %d users", len(userList.Items))
```
choufraise, line 513: **nit**: From an ops perspective, I believe it would be nice to see all blocking resources, like `Found 5 blocking resources: imports-from-system-a, imports-from-system-b, exports-x, user-a, user-b` (sorted alphabetically of course to ensure consistent order)

choufraise (approved):
LGTM

internal/adapter/inbound/controller/account.go
henriropp: sure. I'll save it for later improvement.

PR #270, opened 2026-05-07
test(e2e): replace shell scripts with e2e-ctl

choufraise:
## Summary

Replace shell-heavy KUTTL e2e steps with `e2e-ctl`, a Go-based command binary used directly from test steps.

This makes the e2e flow easier to follow in CI logs, reduces bash-specific behavior, and keeps the test commands in typed Go code where they can be unit tested, linted, and built before KUTTL runs.

## What changed

- Added `test/e2e-ctl` with Cobra commands for:
  - Account operations
  - Account JWT upload
  - e2e assertions
  - JetStream stream create/delete
- Updated KUTTL YAML steps to call `e2e-ctl` instead of inline shell/script blocks.
- Removed replaced scripts from `test/script`.
- Build `e2e-ctl` before running KUTTL and add `bin` to `PATH`.
- Include `test/e2e-ctl/**` in regular lint/test PR checks.
- Preserve original KUTTL YAML file paths to keep test-case history readable.

Co-authored-by: OpenAI <noreply@openai.com>


.golangci.yml
```diff
@@ -35,6 +35,9 @@ linters:
           - dupl
           - lll
         path: internal/*
+      - linters:
+          - lll
```
choufraise, line 39: **note**: For test code, like the `e2e-ctl`, it's simply annoying to have to stick to the 120 chars lines. Hence ignoring `lll`.

kejne (approved):
**praise:** It might be more code like this, but it is for sure more maintainable with tests. 🤖 have a much easier time to not screw up your test env.

henriropp (approved):
Nice! As a side effect I am guessing the `e2e-ctl` can also be used in local manual smoke tests 😃

PR #271, opened 2026-05-07
ci(lint): pin golangci-lint from a shared tool version file

### LINKED ISSUE #189: ci: pin golangci-lint version in GitHub Actions to match local linting
## Problem

The GitHub Actions lint step and the local lint workflow are not guaranteed to use the same `golangci-lint` version.

Locally, `make lint` installs and runs the version pinned in the `Makefile`.

In CI, the workflow uses `golangci/golangci-lint-action` without explicitly pinning the `golangci-lint` version.

This can lead to confusing situations where:
- `make lint` passes locally
- the PR lint check fails in GitHub Actions
- the difference is caused by tool-version drift rather than the code itself

## Why this matters

Linting should be deterministic and reproducible.

If local and CI lint runs do not use the same version, developers cannot reliably verify the result before pushing. That creates unnecessary churn, wastes CI
cycles, and makes the feedback loop slower and less trustworthy.

## Proposal

Pin the `golangci-lint` version explicitly in the GitHub Actions workflow to the same version used by the local `Makefile`.

The goal is that:
- `make lint` locally
- and the PR lint step in GitHub Actions

produce the same result for the same revision.

## Acceptance criteria

- the GitHub Actions lint workflow explicitly pins a `golangci-lint` version
- the pinned CI version matches the version used by `make lint`
- running lint locally and in CI produces consistent results for the same code

## Notes

This is not about changing lint rules. It is about making the lint environment reproducible and aligned between local development and CI.

choufraise:
Add .tool-versions as the single source of truth for the golangci-lint version and use it from the GitHub Actions lint workflow.

Update the Makefile to install the same pinned golangci-lint version locally and switch to the v2 module path.

Remove the duplicate golangci-lint pin from mise.toml so local tooling, Makefile linting, and CI all resolve the same version.

Closes: #189


henriropp (approved):
Nice! Can we finally catch all lint errors before pushing?

PR #272, opened 2026-05-07
test: simplify account import/export KUTTL suites and add deterministic conflict coverage

henriropp:
## Description
Refactors the account import/export KUTTL suites to make them easier to read without reducing coverage.

Stable status and claim assertions are now moved into plain YAML wherever possible. CEL is kept only for genuinely dynamic relationships, such as:

- resolved account IDs
- adopted UIDs
- generation comparisons

**Sorting**
This branch also makes conflict behavior deterministic by sorting adopted import/export resources before claim building. As a result, the new account-import-conflict suite and the updated account-export-conflict suite can assert winner/loser conditions directly in YAML instead of relying on complex exclusivity CEL.

READMEs deleted.


internal/adapter/inbound/controller/account_export.go
```diff
@@ -173,6 +173,9 @@ func (r *AccountExportReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{RequeueAfter: requeueImmediately}, nil
 	}
 
+	// sort conditions before save (helps keep kuttl tests robust)
```
choufraise: **nit**: Instead of mentioning kuttl tests here we could refer to consistency, i.e. that we produce a predictable order.

choufraise (dismissed):
Very nice! :shipit:

choufraise (approved):

choufraise (approved):

PR #290, opened 2026-05-08
feat(natscluster): block deletion while accounts are bound

henriropp:
## Summary

Prevent `NatsCluster` deletion from completing while `Account` resources are still bound to that cluster. This protects cluster/account relationships from being orphaned and surfaces the blocked deletion state through controller status until the bindings are removed.

## What changed

- add a `NatsCluster` finalizer and deletion logic that checks for bound accounts by cluster UID before allowing removal
- watch `Account` delete events from the `NatsCluster` controller so deletion is retried after bindings are removed
- persist the bound cluster UID on `Account` resources via `account.nauth.io/nats-cluster-id`
- reject rebinding an account to a different cluster UID once it has already been bound
- update manager RBAC to allow `natsclusters` updates/finalizer updates and add Helm test coverage for those permissions
- extend KUTTL coverage to verify account-to-cluster binding and add an e2e scenario that confirms `NatsCluster` deletion is blocked until the bound account is deleted
- refactor account and nats cluster controller tests to setup state in given, without using the reconcile function that it is testing
- refactor cluster target lookup to happen in controller, and pass down to manager.

## How to validate

```bash
make lint
make test
make test-e2e

### CODE REVIEW

File: test/e2e/cluster-ref-test/02-assert.yaml [hunk 1]
```diff
@@ -36,8 +36,15 @@ resourceRefs:
...
+  - apiVersion: nauth.io/v1alpha1
+    kind: NatsCluster
+    name: test-cluster
+    namespace: cluster-ref-test
+    ref: cluster
 assertAll:
   # Verify account ID label is set (starts with A)
   - celExpr: matches(account.metadata.labels["account.nauth.io/id"], "^A.{55}$")
+  ## Verify account is bound to nats cluster
+  - celExpr: matches(account.metadata.labels["account.nauth.io/nats-cluster-id"], cluster.metadata.uid)
```
choufraise, line 48: **praise**: Awesome check! 👍

internal/adapter/inbound/controller/natscluster_test.go
```diff
@@ -145,61 +209,69 @@ func (t *NatsClusterControllerTestSuite) Test_Reconcile_ShouldFail_WhenValidatio
...
-		t.Equal(conditionReasonErrored, c.Reason)
-	}
-	t.Len(t.fakeRecorder.Events, 1)
+
+	c := meta.FindStatusCondition(cluster.Status.Conditions, conditionTypeReady)
+	t.Equal(metav1.ConditionFalse, c.Status)
+	t.Equal(conditionReasonErrored, c.Reason)
+
+	t.Require().Len(t.fakeRecorder.Events, 1)
 	t.Contains(<-t.fakeRecorder.Events, "failed to resolve NatsCluster target: a test error")
```
choufraise, line 251: **check**: I know we "fixed" the timeout issue by adding `Require().Len(...)` before this, but can't we improve this check with an explicit timeout of e.g. 2 seconds instead of getting hit by the 10 minutes timeout?

choufraise (dismissed):
👍 LGTM

choufraise (approved):

PR #293, opened 2026-05-11
ci: support release candidate tags

choufraise:
We need to be able to validate major changes before publishing a stable release. RC tags let us ship full testable artifacts without making a known-risk version the default upgrade target or fragmenting final release notes.


henriropp (approved):
LGTM

PR #294, opened 2026-05-11
fix: align monitor alerts with controller metrics

choufraise:
The observability guide documents the active controller labels, but the chart still alerted on a stale operator controller and missed newer controllers. Render alerts for the same controller set exposed by controller-runtime metrics.


henriropp (approved):
LGTM

Issue #300, opened 2026-05-13
Review and consider need of events in reconcilers


henriropp:
**Is your feature request related to a problem? Please describe.**
When something goes wrong, temporary or longer time, we need to be able to quickly resolve the issue. Having good logging/status/events might be necessary.

**Describe the solution you'd like**
Consider if adding events in the reconcilers would help in troubleshooting when troubles arise. 

**Describe alternatives you've considered**
Export and import have multiple conditions that fullfil part of this need.

PR #303, opened 2026-05-18
chore: cleanup todos

henriropp:
Cleaning up remaining todos related to import/export.

### DIFF
--- a/internal/adapter/inbound/controller/account_export.go
+++ b/internal/adapter/inbound/controller/account_export.go
@@ -65,8 +65,6 @@ func NewAccountExportReconciler(k8sClient client.Client, scheme *runtime.Scheme,
 func (r *AccountExportReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := logf.FromContext(ctx)
 
-	// TODO: #11 Consider adding events
-
 	state := &v1alpha1.AccountExport{}
 	if err := r.Get(ctx, req.NamespacedName, state); err != nil {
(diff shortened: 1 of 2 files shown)


internal/adapter/inbound/controller/account_export.go
```diff
@@ -65,8 +65,6 @@ func NewAccountExportReconciler(k8sClient client.Client, scheme *runtime.Scheme,
 func (r *AccountExportReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := logf.FromContext(ctx)
 
-	// TODO: #11 Consider adding events
```
henriropp, line 68: Created Issue
https://github.com/WirelessCar/nauth/issues/300

internal/adapter/inbound/controller/account_import.go
```diff
@@ -316,8 +316,6 @@ func (r *AccountImportReconciler) mapExportAccountToAccountImports(ctx context.C
 
 	imports := &v1alpha1.AccountImportList{}
 	if err := r.List(ctx, imports,
-		// TODO: [#11] handle case when Nauth is bound to single namespace
```
henriropp, line 319: According to AI, when the operator is in single (or default namespaces) mode, the cache, that the `List()` uses if effectively only populated with resources from that/those namespaces. 

In other words, if operator is running in single namespace mode, the `List()` will only search in that namespace.

No special  code needed.

choufraise (approved):
LGTM

Issue #140, opened 2026-02-24
Support AccountSigningKey resources for scoped signing keys and implicit user authorization


choufraise:
## Summary

Introduce an `AccountSigningKey` child resource for `Account` to support NATS scoped signing keys.

This would give NAuth a first-class way to manage account signing keys, including rotation, and allow `User` resources to reference a signing key resource instead of carrying all authorization directly in the user JWT.

Related:
- #135
- #95
- [NATS scoped signing keys](https://docs.nats.io/using-nats/nats-tools/nsc/signing_keys)

## Problem

Today user authorization is primarily expressed explicitly in the `User` JWT.

That creates two problems:

1. authorization changes require re-issuing user JWTs
2. account signing key lifecycle is not modeled as a managed resource in NAuth

With the introduction of `AccountExport`, we now have a clear pattern for account child resources. Signing keys should follow the same direction.

NATS scoped signing keys are the right primitive here:
- authorization can be bound to the account signing key
- the user JWT can stay focused on identity
- signing keys can be rotated independently
- operational changes become faster and safer

## Proposed solution

Add a new `AccountSigningKey` CRD as an `Account` child resource.

The resource should represent an account signing key and its effective scope/permissions.

The `Account` should adopt `AccountSigningKey` resources as input when building the aggregated account JWT.

A `User` should then be able to reference an `AccountSigningKey` resource. When doing so, the resulting user JWT should inherit authorization implicitly from the scoped signing key rather than embedding the same permissions explicitly in the user JWT.

In other words:

- `AccountSigningKey` defines authorization scope
- `Account` adopts `AccountSigningKey` resources into the aggregated account JWT
- `User` references `AccountSigningKey`
- issued user JWT is signed through that scoped signing key
- authorization is derived from the signing key, not duplicated in every user JWT

## Why this is better

This separates identity from authorization.

Benefits:
- faster permission updates
- less JWT churn for users
- cleaner authorization model
- first-class support for signing key rotation
- better alignment with native NATS scoped signing key behavior

It also gives NAuth a path to manage signing key rotation as part of the account lifecycle instead of treating signing keys as opaque implementation details.

## Scope

Implement support for:

1. an `AccountSigningKey` CRD as input to `Account`
2. expressing scoped permissions/claims on the signing key resource
3. adopting `AccountSigningKey` resources into the aggregated account JWT
4. automatic or managed rotation of account signing keys
5. referencing an `AccountSigningKey` from `User`
[...]

Issue #95, opened 2026-01-06
Deleting a User should revoke it


BZValoche:
When deleting a User, it's credentials remain valid for the lifetime of the NATS account.

The credentials should be revoked, please see: https://docs.nats.io/using-nats/nats-tools/nsc/revocation

kejne:
I see that we we're missing an issue that was actually part of the backlog: #132 
The intention is that the signing key is rotated automatically periodically.

This in combination with adding users to revocation lists would solve the issue of the deleting users.
I would prefer not to add additional signing keys, but keep it to signing keys that are rotated and handled by nauth and that specific users are handled by revocation list instead. (The revocation list can then be cleared when the key is rotated).

My suggestion would be to first add the functionality of adding User to revocation list if deleted. Then at a later stage we would be able to add functionality to reduce the revocation list during rotation.

Regarding the patterns where the User controller returns an error: This is intentional and is how the controller pattern works. If it it cannot reconcile, it returns an error in order to try again later. Not returning an error = reconciliation completed, which is not the case.

PR #306, opened 2026-05-25
feat(account-signing-key): separate signing key lifecycle from account

### LINKED ISSUE #135: Separate signing key lifecycle from account
There are different use cases for using signing keys. They can be used to reduce the blast radius of users & exports. Signing keys should be possible to rotate without effort, ideally under the hood.

The current setup with one signing key (with possibility to rotate it automatically) can be a bit too broad depending on use case. For example if private exports are to be added as functionality where different accounts can get different subjects assigned in their token.

## Suggestion
Create a new `AccountSigningKey` CRD which keeps the references to underlying signing keys. It holds information to be used when rotating signing keys and can be referenced when signing users or activation tokens.
For a simple account, one signing key is enough. For more complex accounts, the team can setup arbitrarily complex signing keys constellations depending on use case.

### LINKED ISSUE #138: Support predictable secret names for external consumers
When integrating with external systems that issue NATS User JWTs (e.g., authorization services in a decentralized JWT schema), those systems need access to the Account Signing Key secret.

Following #44, NAUTH names secrets with a hash suffix derived from the account's public key:
```
<account>-ac-sign-<hash>
```

This solved #9 where orphaned account IDs could cause new accounts to overwrite existing secrets, leading to credential loss.

However, it makes the secret name unpredictable until the Account resource is created and reconciled. This makes it hard to reference the secret in a Helm chart or GitOps.

The deprecated naming templates (`<account>-ac-root`, `<account>-ac-sign`) would work for this use case, but, apart from being scheduled for removal #102, there is no (declarative) way to opt into the old naming for new accounts.

I suggest allowing user to specify custom/predictable secret names, e.g.
```yaml
spec:
  secrets:
    rootName: "my-account-root"
    signName: "my-account-sign"
```
Users could set one, both, or neither (defaulting to hash-based naming for unspecified ones).

To mitigate #9, which motivated moving away from predictable secret names, if a secret with that name already exists, fail reconciliation with a clear error until a user manually deletes the existing secret.

Let me know if this approach works. I can try to pick this up if time allows.

wingsofovnia:
## Summary

Adds a first-class `AccountSigningKey` CRD and wires it into `Account`/`User` reconciliation.

Follows the CRD-based direction discussed in #135 and #140.

### AccountSigningKey CRD
  - Standalone, coupled with neither Account nor User lifecycle. Manages or observes account signing-key material.
  - Generates an account signing key seed and stores it in a Secret. 
  - Supports "observe" mode (`nauth.io/management-policy=observe`) just like `Account` CRD - reads an existing Secret without modifying it.
  - `spec.secretName` allows a predictable Secret name (#138).  Required in observe mode. In managed mode it is optional and defaults to `<resourceName>-ac-sign`.
  - Removal of the generated managed Secret is handled by GC via `metadata.ownerReferences`.
- `signingKeyRefs` is a list of `AccountSigningKeyRef` structs (`{kind, name}`), leaving room for future kinds (e.g. issuers) and cross-namespace references without a breaking API change. Modelled after `URLFromReference` in `nats_cluster_types.go`. See [#306 (comment)](https://github.com/WirelessCar/nauth/pull/306#discussion_r3371948529).

### Account trust
  - Added `Account.spec.signingKeyRefs`.
  - Referenced `AccountSigningKey` resources are resolved by the Account reconciler.
  - Their public keys are added to the Account JWT signing-key trust set.
  - The existing implicit/default Account signing key is preserved.

### User signing
- Added `User.spec.signingKeyRef`.
- When set, the User JWT is signed with the referenced `AccountSigningKey`.
- When unset, User reconciliation falls back to the existing implicit Account signing key.
- User reconciliation verifies that the referenced signing key is both desired and published as trusted by the Account before issuing credentials:
 - `User.spec.signingKeyRef` must be listed in `Account.spec.signingKeyRefs`
 - the referenced `AccountSigningKey` must be Ready
 - its public key must appear in `Account.status.claims.signingKeys`
- User reconciliation watches Account signing-key trust changes for prompt requeues.
- `signingKeyRef` follows the same `AccountSigningKeyRef` struct shape as `Account.spec.signingKeyRefs`.

```yaml
# Account
signingKeyRefs:
  - name: app-auth-v1
# User
signingKeyRef:
  name: app-auth-v1
```

### Internal refactoring

- `UserManager` now follows the same style as `AccountManager`: it receives a domain-level `nauth.UserRequest` and returns a `nauth.UserResult`.
- `userClaimsBuilder` was reshaped into a proper builder, matching the style of `accountClaimsBuilder`.
- User creds Secrets are now cleaned up through owner references instead of explicit finalizer deletion.
  - The User finalizer is kept for now because it is still a useful hook for future User revocation work (#95, NATS revocation support).
- Secret handling now uses controller owner references and exposes ownership checks so managed resources can reject unrelated existing Secrets safely.

### Tests

Three new KUTTL e2e harnesses cover the trust chain end-to-end against a live NATS:
- `test/e2e/account-signing-key` - happy path: create `AccountSigningKey`, trust it on Account, sign a User with it.
- `test/e2e/account-signing-key-rotation` - overlap-rotation flow (v1 -> both -> v2 -> retire v1).
- `test/e2e/account-signing-key-revocation` - remove key from `Account.spec.signingKeyRefs` and assert User is marked NotReady.

## Signing Key Rotation

Signing-key rotation is modeled as replace-by-resource, not in-place Secret mutation. A typical rotation flow is:

1. Create a new `AccountSigningKey`, for example `app-auth-v2`.
2. Add it to `Account.spec.signingKeyRefs` alongside the old key.
3. Account reconciliation publishes an Account JWT that trusts both old and new signing keys.
4. Patch Users or external consumers from the old signing key to the new one.
5. Remove the old key from `Account.spec.signingKeyRefs`.
6. After a grace period, delete the old `AccountSigningKey`.

The implicit/default Account signing key remains available for Users that do not specify `spec.signingKeyRef`. Whether implicit keys stay in the longer term or we should go `AccountSigningKey` all the way is TBD.

Closes #135. Closes #138. Partially advances #140 (no scoped signing-key permissions and implicit authorization).


choufraise (commented):
Thanks for the contribution. 💪  The PR is well split for review, and I appreciate the focused commits and test coverage. I also appreciate that this aligns the User context with the Account-side refactorings we have done over the last months. We have not spent much time on the User parts yet, so that cleanup is valuable.

Before I go into the implementation details, I would like to align on two broader points.

First, can you share how you see this design evolving toward NAuth-managed signing-key rotation?

My reading is that the current API models explicit key replacement: create a new `AccountSigningKey`, add both keys to the `Account`, update every `User` to the new key, then remove the old key. That works, but it puts the actual rotation workflow outside NAuth, typically in GitOps/automation.

There are several rotation scenarios we should design for:
- scheduled rotation driven by security/compliance requirements, for example every 24h or every 30 days
- emergency rotation after a leaked signing key or leaked user credentials
- operational rotation before handing ownership between teams/environments
- preventive rotation after restoring from backup or suspected cluster/Secret exposure

Because of that, I think the rotation workflow needs to be simple from the user perspective. Ideally NAuth should support both policy-driven rotation and an explicit “rotate now” trigger. One possible shape is a stable signer/key-ring concept with active + previous/next keys, an overlap period, and something like `spec.rotation.requestId` for manual rotation requests, with NAuth recording the handled request in status.

I am not saying this PR must implement all of that, but I would like us to discuss whether the current `AccountSigningKey -> one immutable Secret` model leaves that path open.

Second, please autosquash the three `fix:` commits into the commits that introduced the affected code. This is not about reducing the number of commits. Small commits are good. These are fixes to new code in this PR, not fixes to existing released behavior.

Suggested mapping:
- `048f059` and `b3f8f2e` -> `a20eda8`
- `2b9320b` -> `8a8e395`

wingsofovnia:
> First, can you share how you see this design evolving toward NAuth-managed signing-key rotation?
>
> My reading is that the current API models explicit key replacement: create a new `AccountSigningKey`, add both keys to the `Account`, update every `User` to the new key, then remove the old key. That works, but it puts the actual rotation workflow outside NAuth, typically in GitOps/automation.
> 
> There are several rotation scenarios we should design for:
> * scheduled rotation driven by security/compliance requirements, for example every 24h or every 30 days
> * emergency rotation after a leaked signing key or leaked user credentials
> * operational rotation before handing ownership between teams/environments
> * preventive rotation after restoring from backup or suspected cluster/Secret exposure
> 
> Because of that, I think the rotation workflow needs to be simple from the user perspective. Ideally NAuth should support both policy-driven rotation and an explicit “rotate now” trigger. One possible shape is a stable signer/key-ring concept with active + previous/next keys, an overlap period, and something like spec.rotation.requestId for manual rotation requests, with NAuth recording the handled request in status.
>
> I am not saying this PR must implement all of that, but I would like us to discuss whether the current AccountSigningKey -> one immutable Secret model leaves that path open.

Your reading of the current API model is correct.

I think `AccountSigningKey` might work well as a leaf resource: one resource represents one concrete signing key and one Secret. It supports explicit replacement today and gives a stable building block for higher-level key management later.

Taking inspiration from cert-manager, there could be an `AccountSigningKeyIssuer` governing a class of signing keys that both `User` and `Account` can reference without being tied to individual keys.

```yaml
apiVersion: nauth.io/v1alpha1
kind: AccountSigningKeyIssuer
metadata:
  name: sketchy
spec:
  accountName: my-account
  rotation:
    duration: 2160h
    renewBefore: 360h
    rollover: 12h # Keep old key trusted for an overlap period.
```

This would loosely (except the overlap period) keep 1:1 mapping to a `AccountSigningKey`. 

The issuer controller could either perform scheduled rotation directly, or later create a separate `AccountSigningKeyRotationRequest` if we want each rotation to have its own auditable lifecycle object and to make it easier to perform ad-hoc / emergency rotations:

```yaml
 apiVersion: nauth.io/v1alpha1
 kind: AccountSigningKeyRotationRequest
 metadata:
   name: sketchy-scheduled-rotation-20260526
   ownerReferences:
     - kind: AccountSigningKeyIssuer
       name: sketchy
 spec:
   issuers:
     - name: sketchy
   reason: Scheduled
   mode: Graceful # Keep old key trusted for an overlap period.
 ```
 
 For emergency or ad-hoc rotation, users would create the same request explicitly:

 ```yaml
 apiVersion: nauth.io/v1alpha1
 kind: AccountSigningKeyRotationRequest
 metadata:
   name: leaked-keys-april-2026
 spec:
   issuers:
     - name: sketchy
     - name: prod
   reason: Emergency
   mode: Immediate # Remove old key(s) from Account trust immediately.
```

`AccountSigningKeyRotationRequest`'s controller will likely be the one performing actual rotation.

A `User` then can reference the issuer:
```yaml
apiVersion: nauth.io/v1alpha1
kind: User
metadata:
  name: user-for-doing-sketchy-stuff
spec:
  accountName: my-account
  signingKeyIssuer: sketchy
```

The `User` reconciler could reissue credentials when the issuer’s active key changes. `Account` would have a similar `signingKeyIssuers` field to trust keys issued by one or more issuers.

This seems like the cleanest option to me, though the tradeoff is additional CRD/API surface.

That said, immutable AccountSigningKey appears to fit this model well. Account.spec.signingKeyRefs and User.spec.signingKeyRef can remain useful for explicit, self-managed keys even when automated key management is introduced later.

Would be great to hear your thoughts on this!

> Second, please autosquash the three fix: commits

~Anticipating further commits, I'd say this can come last once we agree on the direction in general and I will get things right. Would that be alright?~ Squashed.

BZValoche:
Why not annotate the `AccountSigningKeyIssuer` for emergency or ad-hoc rotations?

For example:
```bash
kubectl annotate AccountSigningKeyIssuer sketchy nauth.io/rotate=immediate
kubectl annotate AccountSigningKeyIssuer sketchy nauth.io/rotate=graceful
```
And also, scheduled, automatic key rotations should not be mandatory IMO.

wingsofovnia:
> And also, scheduled, automatic key rotations should not be mandatory IMO - @BZValoche 

I agree. The way I put `AccountSigningKeyIssuer` in my suggestion was more like a `AccountSigningKeyRotator` or smth so I wasn't really thinking of any other use case to define it but rotation. However, if we took a broader view, `AccountSigningKeyIssuer` could become a replacement to implicit signing key generation _with a rotation feature_. We could deprecate implicit keys and instead advise defining `AccountSigningKeyIssuer`. We could even drop implicit signing key altogether and say if you don't use `AccountSigningKeyIssuer`, users will be signed with account's nkey (which is insecure and you should use `AccountSigningKeyIssuer`).

> Why not annotate the AccountSigningKeyIssuer for emergency or ad-hoc rotations? - @BZValoche 

I find using annotation for this quite confusing in terms of managing state/status. Say we put `nauth.io/rotate=immediate` and it does so on reconciliation. Should it remove its own annotation afterwards? I find it awkward for a controller to modify its definition and GitOps might fight controller-cleared annotations too. It is also harder to validate with CRD schema/CEL and we should re-work controllers reconciliation predicates to get triggered by metadata-only annotation changes. 

With `AccountSigningKeyRotationRequest` on the other hand we have a clear record of rotation, with its own status. It's easy to debug and track.

@choufraise would you or other maintainers will have some time to look into this PR/proposal anytime soon? Thank you.

BZValoche:
> I find using annotation for this quite confusing in terms of managing state/status. Say we put `nauth.io/rotate=immediate` and it does so on reconciliation. Should it remove its own annotation afterwards? I find it awkward for a controller to modify its definition and GitOps might fight controller-cleared annotations too. It is also harder to validate with CRD schema/CEL and we should re-work controllers reconciliation predicates to get triggered by metadata-only annotation changes.

Yes I agree, but if you are deploying using for example ArgoCD, then a manual rotate operation would live forever in git.
Maybe a better pattern would be like this one:
```
nauth.io/rotate-immediate-requested-at: `date +%s`
nauth.io/rotate-graceful-requested-at: `date +%s`
```
Then the operator can process and store the status accordingly:
```yaml
status:
  rotateImmediateRequestedAt: XXXXXX
  lastRotateImmediateCompletedAt: XXXXXX
```
Then the operator can compare `annotation.rotate-xxx-requested-at > status.lastRotateXxxCompletedAt`.

But it is not that important, just a use-case that I can foresee :
Most likely, immediate manual rotation are emergency measures, one might not want to go through a complex git process (issue + MR + approval + merge, not mentionning CI pipelines which could take a while).

Having a ressource corresponding to a time bound action seems a bit odd to my ears in a declarative context.

Anyway I could debate this subject during hours, so I won't bother you more. The approval of this PR is much awaited from my side, thank you so much for your job 👍

api/v1alpha1/account_types.go
```diff
@@ -67,6 +64,16 @@ type AccountSpec struct {
...
+	// SigningKeyRefs lists AccountSigningKey resource names in the same namespace
+	// whose public keys are trusted as additional signing keys for this account.
+	// The implicit default signing key is always present and is not listed here.
+	// +optional
+	// +listType=set
+	// +kubebuilder:validation:MaxItems=64
+	// +kubebuilder:validation:items:MinLength=1
+	// +kubebuilder:validation:items:MaxLength=253
+	// +kubebuilder:validation:items:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+	SigningKeyRefs []string `json:"signingKeyRefs,omitempty"`
```
choufraise: **suggestion**: When thinking about Auth Callout support, such a service would have to sign temporary users using an Account signing key. The trust to such a service should be on the Account level, i.e. the Account trusts the "Auth Callout Service `AccountSigningKeyIssuer`" for instance. In a multi-namespaced cluster you could also have a trusted service that deals with signing keys for all Accounts, cross namespace. I'm thinking that maybe we should introduce a struct, e.g. `AccountSigningKeyRef`, for this instead of `[]string` to enable extensibility towards issuers and "raw" signing keys in other namespaces, but also to enable different resource kinds. Somewhat similar to the `URLFromReference` resource in `nats_cluster_types.go`. This would ensure future compatibility with whatever _other_ signing key resources we come up with.  

This of course also applies to the `User.spec.SigningKeyRef` field.

choufraise (commented):
Sorry for a late response, it's been some hectic couple of weeks. 

As mentioned, my main focus is still on the extensibility and future compatibility on the API/CRDs. See my comment.

Also...

> We could even drop implicit signing key altogether and say if you don't use AccountSigningKeyIssuer, users will be signed with account's nkey (which is insecure and you should use AccountSigningKeyIssuer).

This is one thing I highly believe we should avoid at all cost. Signing even a single User using the Account Root key would grant that user access until actively revoked or until `exp` expires, but since the default today is to not even set an expiration date on the User JWT those credentials will be valid during the entire lifespan of the account.

api/v1alpha1/account_types.go
wingsofovnia: Added in [6f6c49c](https://github.com/WirelessCar/nauth/pull/306/commits/6f6c49ca523b13f2ca2545c51a63ef0ab1faf50d)

wingsofovnia:
Should be good to go again. Hope this covers all the bases. It becomes quite hard for me to keep up with merge conflicts around the user reconciler / claims refactor, so anything that could be a follow-up I'm glad to file as a separate issue/PR.

@choufraise, would you have time to take another look this week and see if we can land it? Thank you!

internal/adapter/outbound/k8s/secret.go
```diff
@@ -56,8 +55,10 @@ func (k *SecretClient) Apply(ctx context.Context, owner metav1.Object, meta meta
 		maps.Insert(currentSecret.Labels, maps.All(meta.Labels))
 
 		currentSecret.StringData = valueMap
-		if err := addOwnerReferenceIfNotExists(currentSecret, owner); err != nil {
-			return err
+		if owner != nil {
+			if err := controllerutil.SetControllerReference(owner, currentSecret, k.client.Scheme()); err != nil {
```
choufraise, line 59: **praise**: Nice catch!

internal/core/account.go
```diff
@@ -77,9 +77,9 @@ func (a *AccountManager) CreateOrUpdate(ctx context.Context, request nauth.Accou
 
 	cluster := request.ClusterTarget
 	fixedAccountID := string(request.AccountID)
+
 	accountSecrets, found, err := a.secretManager.GetSecrets(ctx, request.AccountRef, fixedAccountID)
 	if fixedAccountID != "" {
-		// Update
```
choufraise, line 82: **question**: Why did you feel the urge to remove the `// Update` and `// Create` comments?

internal/adapter/inbound/controller/user.go
```diff
@@ -84,40 +98,40 @@ func (r *UserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
-			log.Info("Failed to update the user status", "name", user.Name, "error", err)
-			return ctrl.Result{}, err
-		}
-
-		// The user is being deleted
-		if controllerutil.ContainsFinalizer(user, finalizerUser) {
-			if err := r.manager.Delete(ctx, user); err != nil {
-				return r.reporter.error(ctx, user, fmt.Errorf("failed to delete user: %w", err))
-			}
+		// TODO: r.manager.Revoke(UserRevocationRequest) to revoke the user from the Account, see
```
choufraise: **nit**: We normally reference `TODO` issues using this format: `// TODO: [#95]` since it simplifies search. IDEs normally also displays only the first line in built-in TODO-lists.

internal/adapter/inbound/controller/user.go
```diff
@@ -170,11 +210,257 @@ func (r *UserReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+	for _, ref := range account.Spec.SigningKeyRefs {
+		if ref.Name == refName {
+			refInSpec = true
+			break
+		}
+	}
+	if !refInSpec {
+		return nil, fmt.Errorf("signing key ref %q is not listed in Account %q spec.signingKeyRefs: %w",
+			refName, user.Spec.AccountName, errDependencyNotReady)
+	}
```
choufraise, line 302: **nit**: With the introduction of `AccountExport` and `AccountImport` CRDs we've tried to put the business logic in respective controller and/or manager. In this case, the `User` controller assumes that a signing key is trusted by looking at the referenced `Account` _spec_. Now, the `Account` may not even be `Ready`, meaning the `spec` is yet to be validated and accepted. Even if we would also check that the `Account` is ready, it would still be an assumption from an external party. I'd prefer it if we would put such business login into the `AccountManager` to ensure they use exactly the same mechanism to determine if a signing key is valid, trusted and ready to be used.  Currently, the only way to verify that this assumption is true is by covering those use-cases with e2e tests. 

Please note, I don't expect you to fix this as it's (really) good enough for now, but since the project is growing quite rapidly and with a lot of new planned CRDs coming in, we should stay away from assumptions imo.

api/v1alpha1/account_types.go
```diff
@@ -67,6 +64,41 @@ type AccountSpec struct {
...
+// AccountSigningKeyRefKind is the kind of resource referenced as an account
+// signing key.
+// +kubebuilder:validation:Enum=AccountSigningKey
+type AccountSigningKeyRefKind string
+
+const (
+	AccountSigningKeyRefKindAccountSigningKey AccountSigningKeyRefKind = "AccountSigningKey"
+)
+
+// AccountSigningKeyRef references a signing-key resource in the same namespace
```
choufraise: **suggestion**: For major systems, where e.g. you have a general Auth Callout service with its own `AccountSigningKey` (or `AccountSigningKeyProvider` in the future), this resource may be put in a different namespace. Suggestion is to support `namespace` in the reference as well so that the `Account` spec trusts such Auth Callout signing key instead of the other way around.

choufraise (dismissed):
It looks really good! I have some minor comments, nothing you need to take of now, but I'd appreciate your view and input on them. 

Also, thank you for your patience in regard to the lack of https://github.com/WirelessCar/nauth/issues/331.

henriropp (dismissed):

henriropp:
The Signing key rotation process described is only manual?

I see account has watch on AccountSigningKey, and user on account, but the only thing that can be changed is the secretname. SecretName is immutable, so I guess the watch is only in case the resources are created out of order.

If we wanted to automate the signing key rotation, what options do we have? is it possible to replace the secret in place and trigger (or implement watch on it from account)? Maybe replacing, is not an option, as it will render existing accounts/users unusable until they reconcile? 

Automation might be a bit cumbersome, when all resources are GitOpsed, but maybe there is a path forward with defining two AccountSigningKeys, trusting both, and rotating one at the time?

choufraise:
@wingsofovnia , awaiting your input on some conversations and the questions from @henriropp

internal/core/account.go
wingsofovnia: can't remember at this point :) probably by mistake. I'll revert them back with a fixup.

internal/adapter/inbound/controller/user.go
wingsofovnia: Updated with a fixup.

api/v1alpha1/account_types.go
wingsofovnia: Added in c670078

wingsofovnia:
Hi @henriropp, yes, rotation here is manual by design. `AccountSigningKey` is intentionally a leaf resource (one resource = one immutable key + Secret). In-place Secret replacement would invalidate every User JWT already signed with the old key (the new public key wouldn't be in the Account's trust set), which is why `secretName` is immutable. For automated rotation I sketched a higher-level `AccountSigningKeyIssuer` in the earlier thread with @choufraise. It would orchestrate these leaf `AccountSigningKey` resources on a schedule (or on demand via an `AccountSigningKeyRotationRequest`), keeping immutability at the leaf level. That's intended as a follow-up so this PR stays focused.

On the watches: it's not only out-of-order creation. They also react to (1) `AccountSigningKey` going Ready / publishing its public key, (2) entries being added or removed from `Account.spec.signingKeyRefs`, and (3) `Account.status.claims.signingKeys` changing, so the User reconciler waits until the ASK's public key actually appears there before issuing creds. That's what makes the dual-key overlap rotation you described work: add v2 to `signingKeyRefs`, the watches propagate trust to Users, migrate them to v2, then remove v1. Covered end-to-end by the `account-signing-key-rotation` e2e test in this PR.

internal/adapter/inbound/controller/user.go
wingsofovnia: I am ready to do a follow-up on that. Should I create a ticket?

internal/adapter/inbound/controller/account.go
```diff
@@ -386,6 +421,35 @@ func (r *AccountReconciler) findImportsByAccountID(ctx context.Context, namespac
...
+	if len(refs) == 0 {
+		return nil, nil
+	}
+	publicKeys := make([]string, 0, len(refs))
+	for _, ref := range refs {
+		askNamespace := ref.Namespace
+		if askNamespace == "" {
+			askNamespace = namespace
+		}
+		ask := &v1alpha1.AccountSigningKey{}
```
choufraise, line 429: **issue**: Since you introduced `Kind` in the `AccountSigningKeyRef` we need to check that it's either empty or `AccountSigningKey`, anything else must generate a reconcile error due to invalid input. We strive at failing fast with strict input validation.

internal/adapter/inbound/controller/account_test.go
```diff
@@ -688,6 +688,292 @@ func (t *AccountControllerTestSuite) createExport(namespace domain.Namespace, na
...
+
+	account := &v1alpha1.Account{}
+	t.Require().NoError(k8sClient.Get(t.ctx, t.accountNamespacedRef, account))
+	c := meta.FindStatusCondition(account.Status.Conditions, conditionTypeReady)
+	t.Require().NotNil(c)
+	t.Equal(metav1.ConditionFalse, c.Status)
+	t.Equal(conditionReasonReconciling, c.Reason)
+}
+
+func (t *AccountControllerTestSuite) Test_Reconcile_ShouldSetReconciling_WhenSigningKeyReadyButEmptyPublicKey() {
```
choufraise: **question**: When is this ever a use-case? How can `AccountSigningKey` be `Ready` without exposing the public key?

internal/adapter/inbound/controller/account_test.go
```diff
@@ -688,6 +688,292 @@ func (t *AccountControllerTestSuite) createExport(namespace domain.Namespace, na
...
+	account := &v1alpha1.Account{}
+	t.Require().NoError(k8sClient.Get(t.ctx, t.accountNamespacedRef, account))
+	c := meta.FindStatusCondition(account.Status.Conditions, conditionTypeReady)
+	t.Require().NotNil(c)
+	t.Equal(metav1.ConditionFalse, c.Status)
+	t.Equal(conditionReasonErrored, c.Reason)
+	t.Len(t.fakeRecorder.Events, 1)
+}
+
+func (t *AccountControllerTestSuite) Test_Reconcile_ShouldSetReconciling_WhenSigningKeyRefNotReady() {
```
choufraise: **issue**: `AccountImport`- and `AccountExport` resources are today simply ignored by the account reconciliation if they haven't defined a `DesiredClaim`, which I guess would be the equivalent of the Public Key in this case. Suggestion is to treat all sub resources the same way. Any reason why it's not?

choufraise (changes requested):
Found a couple of issues that require your attention @wingsofovnia .

internal/adapter/inbound/controller/account_test.go
wingsofovnia: That state is indeed not reachable by contract (the AccountSigningKey controller writes status.publicKey and the Ready=True condition in the same Status().Update). Test and the matching defensive branch in resolveSigningKeyRefs both removed in dd159c4.

internal/adapter/inbound/controller/account.go
wingsofovnia: Fixed in 4bba20d.

internal/adapter/inbound/controller/account_test.go
wingsofovnia: Fixed in dd159c4. resolveSigningKeyRefs now silently skips missing or not-Ready refs (the AccountSigningKey watch re-enqueues the Account once they become Ready), matching the Imports/Exports model.

choufraise (approved):
Great work! 👌  Thank you a bunch for this contribution. Will create a release including this feature asap so that you can start using it.

internal/adapter/inbound/controller/user.go
choufraise: I created https://github.com/WirelessCar/nauth/issues/336 for this. Resolving conversation to enable PR merge.

choufraise:
@wingsofovnia , release [v0.7.2](https://github.com/WirelessCar/nauth/releases/tag/v0.7.2) created containing this feature. 

I've also created a [CRDs-catalog PR](https://github.com/datreeio/CRDs-catalog/pull/909).

Issue #59, opened 2025-11-21
Add lifecycle policy for nauth resources


kejne:
**Is your feature request related to a problem? Please describe.**
Accounts are critical resources where an unintended deletion could cause a lot of problems with potentially causing cascading effects.

**Describe the solution you'd like**
Tools such as Crossplane use lifecycle policies to manage how resources are managed. For example, the account could take control of an existing account or be set to only add & update.

**Describe alternatives you've considered**
To configure NATS to not allow deletions, which *could* be an option as well.

kejne:
See also Crossplane variants for this:
https://docs.crossplane.io/v2.3/managed-resources/managed-resources/#managementpolicies

PR #308, opened 2026-05-29
feat: add Account lifecycle policy API preview

choufraise:
## Summary

Introduces the Account lifecycle policy API shape for #59.

This adds `spec.lifecycle` with scoped policy fields for:
- NATS Account JWT management/deletion
- Account root/signing Secret management/deletion

Only the default lifecycle is supported at runtime for now. Explicit non-default policies fail fast with `Ready=False` and a clear not-implemented error, so users can see the future API shape without accidentally relying on unsupported behavior.

## Commits

- **feat(api): add Account lifecycle policy fields**
- **feat(controller): reject unsupported Account lifecycle policies**
- **docs: document Account lifecycle preview**

## Notes

- Keeps existing behavior unchanged when `spec.lifecycle` is omitted.
- Keeps legacy observe-mode label behavior unchanged.
- Rejects combining `spec.lifecycle` with the legacy `nauth.io/management-policy: observe` label.
- Adds lifecycle docs and nests observe-mode docs under Account Lifecycle.

Refs #59


api/v1alpha1/account_lifecycle_types.go
```diff
@@ -0,0 +1,76 @@
...
+	AccountSecretsManagementPolicyGenerateIfMissing AccountSecretsManagementPolicy = "GenerateIfMissing"
+	AccountSecretsManagementPolicyRequireExisting   AccountSecretsManagementPolicy = "RequireExisting"
+)
+
+type AccountLifecycleDeletionPolicy string
+
+const (
+	AccountLifecycleDeletionPolicyDelete AccountLifecycleDeletionPolicy = "Delete"
+	AccountLifecycleDeletionPolicyRetain AccountLifecycleDeletionPolicy = "Retain"
+)
```
kejne, line 38: **question:** Can you elaborate on why there are different variations of policies? I believe the crossplane variation is elegant and would fit well for example:
- Observe: nothing created, just fetching secret and getting jwt from nats
- Create: Allows creating new account secret
- Update: Allows updating nats jwt
- Delete: If enabled, deletes account from NATS & secret deleted. Otherwise all are orphaned.

choufraise: The main difference from Crossplane is that NAuth `Account` is not one external resource. It represents two lifecycle surfaces with different risk and behavior, the Acocunt JWT in the NATS cluster and the account secrets.

A flat action list like `Observe`, `Create`, `Update` and `Delete` becomes ambiguous for NAuth.

For example:
- `Create` could mean “generate account Secrets” or “publish an Account JWT to NATS”.
- `Update` could mean “update the NATS Account JWT” or “replace signing key”.
- `Delete` could mean “delete the NATS Account JWT”, “delete the account secrets”, or both.

That distinction matters for DR. The key use case is:

```yaml
natsAccount:
  managementPolicy: Manage
secrets:
  managementPolicy: RequireExisting
```

This means: “manage/update the NATS Account JWT, but do not generate secrets; require restored Secrets.”

A single `Observe` would be too broad here because it would also imply not updating NATS, and `Update` may imply that the account should _not be created_ in the NATS cluster, only updated if already existing.

The split also makes deletion safer. We may want to:
- delete the NATS Account JWT but retain Secrets
- retain the NATS Account JWT but delete Secrets
- retain both on `Account` CR deletion

Those are different lifecycle decisions with different blast radius.

So the design is Crossplane-inspired, but scoped per resource.

This keeps intent explicit, avoids overloaded action names, and leaves room for e.g. future signing-key rotation without abusing `Create` / `Update` semantics.

henriropp (approved):

api/v1alpha1/account_lifecycle_types.go
kejne: I've read through it again now and I would agree that it makes sense for the NAuth use case even though I would still argue the use case of for example deleting the secret but not the NATS account.

kejne (approved):

PR #317, opened 2026-06-02
fix(chart): allow recording events.k8s.io warnings

### LINKED ISSUE #316: Missing events.k8s.io RBAC prevents warning events from being recorded
**Describe the bug**
NAuth cannot record warning Events when a reconciliation error occurs. The controller uses `events.EventRecorder`, which writes Events in the `events.k8s.io` API group, but the Helm chart only grants permissions for core `events`.

This causes noisy controller logs like:

```text
events.events.k8s.io is forbidden: User "system:serviceaccount:nats:nats-nauth" cannot create resource "events" in API group "events.k8s.io"
```

The original reconciliation error is still reported in status, but the Kubernetes Event is rejected.

**To Reproduce**
1. Install NAuth from a version affected by the recorder migration.
2. Trigger any reconciliation error that records a warning Event, for example a User referencing an Account that is not ready.
3. Observe that the controller tries to record the warning Event.
4. Observe an RBAC error for `events.events.k8s.io` creation.

**Versions**
Which versions were used when the problem was observed?
- NAuth (chart & app): affected since `v0.5.2`
- Kubernetes: N/A
- NATS: N/A

**Expected behavior**
NAuth should have RBAC permission to create and patch Events emitted by its controllers.

The manager Role/ClusterRole should include:

```yaml
apiGroups:
- events.k8s.io
resources:
- events
verbs:
- create
- patch
```

**Screenshots**
Not applicable.

**Additional context**
This was introduced by commit `e7f97af50`, which migrated from deprecated `record.EventRecorder` / `GetEventRecorderFor` to `events.EventRecorder` / `GetEventRecorder`. The RBAC stayed on core `events`, which matched the old recorder but not the new one.

choufraise:
Grant the manager Role/ClusterRole permission to create and patch events.k8s.io Events. The controllers use events.EventRecorder, so warning events are rejected without this API group permission.

Keep the existing core events permission for now until the broader RBAC review removes stale permissions deliberately.

Fixes: #316

### DIFF
--- a/charts/nauth/templates/rbac_manager_role.yaml
+++ b/charts/nauth/templates/rbac_manager_role.yaml
@@ -84,3 +84,10 @@ rules:
   verbs:
   - create
   - patch
+- apiGroups:
+  - events.k8s.io
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
--- a/charts/nauth/tests/rbac_manager_role_natscluster_permissions_test.yaml
+++ b/charts/nauth/tests/rbac_manager_role_natscluster_permissions_test.yaml
@@ -1,4 +1,4 @@
-suite: manager role NatsCluster permissions
+suite: manager role permissions
 templates:
   - templates/rbac_manager_role.yaml
 tests:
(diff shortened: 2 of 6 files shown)


henriropp (approved):
LGTM

PR #320, opened 2026-06-03
feat: bootstrap managed Accounts before full reconciliation

### LINKED ISSUE #313: Bootstrap new Accounts with limits before full claim reconciliation
**Is your feature request related to a problem? Please describe.**
New Account creation is blocked when the full Account claims cannot be rendered during the first reconcile. One example is inline imports that reference Accounts whose IDs are not available yet. This slows down disaster recovery and bulk restore flows because the Account is not created in NATS and no Account ID labels/status are written.

**Describe the solution you'd like**
NAuth should bootstrap new Accounts with a minimal Account JWT before reconciling the full desired claims. The bootstrap JWT should include identity/signing material and the supplied limiting fields only:

- `accountLimits`
- `jetStreamEnabled`
- `jetStreamLimits`
- `natsLimits`

Invalid limits must still fail bootstrap. NAuth must not create a temporary limitless Account.

After the base Account exists in NATS and the Account ID labels/status are written, the reconciler should retry the full desired Account JWT. If full claim rendering still fails, the Account should stay `Ready=False` / not ready with a clear error until dependencies or claim conflicts are resolved.

**Describe alternatives you've considered**
Require users to apply Accounts in dependency order or temporarily remove imports/exports during recovery. That works, but it makes disaster recovery and bulk restore slower and more manual.

**Additional context**
This should be treated as a bootstrap behavior for new Accounts only. Existing Accounts should continue to report full reconciliation errors normally.

choufraise:
This adds an initial bootstrap step for managed Accounts that do not yet have an Account ID. The controller now creates a limits-only Account JWT, patches the Account ID and signer labels, and immediately requeues so the next reconcile can perform the full Account reconciliation with imports, exports, adoptions, and status updates.

Also included:
- patch Account labels instead of full resource updates
- update Account Ready status through the Kubernetes helper
- extract Account deletion handling for readability
- keep observed Accounts on the import-only path

Closes: #313

### DIFF
--- a/internal/adapter/inbound/controller/kubernetes.go
+++ b/internal/adapter/inbound/controller/kubernetes.go
@@ -1,0 +1,43 @@
+package controller
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type kubernetesClient struct {
+	client.Client
+}
+
+func newKubernetesClient(k8sClient client.Client) *kubernetesClient {
+	return &kubernetesClient{
+		Client: k8sClient,
+	}
+}
+
+type metadataPatch struct {
+	Metadata labelsPatch `json:"metadata"`
+}
+
+type labelsPatch struct {
+	Labels map[string]string `json:"labels"`
+}
+
+func (c *kubernetesClient) PatchLabels(ctx context.Context, resource client.Object) error {
+	patchData, err := json.Marshal(metadataPatch{
+		Metadata: labelsPatch{
+			Labels: resource.GetLabels(),
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to generate label patch: %w", err)
+	}
+	if err = c.Patch(ctx, resource, client.RawPatch(types.MergePatchType, patchData)); err != nil {
+		return fmt.Errorf("failed to patch labels: %w", err)
+	}
+	return nil
+}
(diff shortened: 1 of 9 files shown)


internal/adapter/inbound/controller/kubernetes.go
```diff
@@ -0,0 +1,62 @@
...
+	})
+	if err := c.Status().Update(ctx, resource); err != nil {
+		return fmt.Errorf("failed to update ready status: %w", err)
+	}
+	return nil
+}
+
+func (c *kubernetesClient) UpdateReadyStatusReconciled(ctx context.Context, resource Object) error {
+	return c.UpdateReadyStatus(ctx, resource, metav1.ConditionTrue, conditionReasonReconciled, "Successfully reconciled")
+}
```
choufraise, line 62: **note**: This is a prep to refactor away from the `statusReporter` concept into something more generic since we have many more reconcilers nowadays it would be nice to keep behaviors and code like this centralized to prevent re-inventing the wheel in each controller.

mise.toml
```diff
@@ -35,7 +35,7 @@ alias = "nld"
 run = "vale README.md www/src/content/docs/index.mdx www/src/content/docs/guides"
 
 [tasks."nauth:e2e-test"]
-silent = true
+silent = false
```
choufraise, line 40: **note**: The e2e tests are long-running and it's extremely annoying to run them without any insight until it's completely done, which could take minutes based on the current cache state etc. Hence, making this non-silent.

henriropp (approved):
Looks nice and clean 🥇

PR #321, opened 2026-06-04
build(mise): add CRDs-catalog update workflow

henriropp:
## Summary
- Add a `mise nauth:update-crds-catalog <catalog_dir>` task for generating NAuth schemas in a forked CRDs-catalog checkout.
- Document the fork-first catalog update workflow.
- Add the CRDs-catalog update to release steps when CRDs change.

## Verification
The workflow has been tested, and a new [PR has been created](https://github.com/datreeio/CRDs-catalog/pull/893) to update the catalog.


kejne (commented):
This is a custom extraction of CRDs which would need to be added in the PR description.

Another way (which was used before) is to use the regular local setup and utilize the CRD extractor tool which they advice.

https://github.com/datreeio/CRDs-catalog/blob/main/.github/pull_request_template.md

If the process includes cloning the repo as a subdirectory in the existing repo this needs to be handled as well.

kejne (approved):

PR #322, opened 2026-06-04
feat(chart): add .Values.extraResources

madestreel:
Goal is to allow users of the helm chart to deploy extra resources along nauth. This is useful to pre-configure a NatsCluster or run the operator-bootstrap within a job.

### DIFF
--- a/charts/nauth/templates/extra_resources.yaml
+++ b/charts/nauth/templates/extra_resources.yaml
@@ -1,0 +1,4 @@
+{{- range .Values.extraResources }}
+---
+{{ toYaml . }}
+{{- end }}
--- a/charts/nauth/templates/extra_resources.yaml
+++ b/charts/nauth/templates/extra_resources.yaml
@@ -1,4 +1,4 @@
 {{- range .Values.extraResources }}
 ---
-{{ toYaml . }}
+{{ tpl (toYaml .) $ }}
 {{- end }}
--- a/charts/nauth/values.yaml
+++ b/charts/nauth/values.yaml
@@ -145,3 +145,5 @@ crds:
   install: true
   # -- Indicates if Custom Resource Definitions should be kept when a release is uninstalled.
   keep: true
+
+extraResources: []
--- a/charts/nauth/values.yaml
+++ b/charts/nauth/values.yaml
@@ -146,4 +146,5 @@ crds:
   # -- Indicates if Custom Resource Definitions should be kept when a release is uninstalled.
   keep: true
 
+# -- Deploy extra resources along the chart. Supports templating
 extraResources: []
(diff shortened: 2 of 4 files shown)


charts/nauth/templates/extra_resources.yaml
```diff
@@ -0,0 +1,4 @@
+{{- range .Values.extraResources }}
+---
+{{ toYaml . }}
```
choufraise: **issue**: `tpl` should be used when rendering `extraResources`, otherwise e.g. `{{ .Release.Name }}` stays literal in user resources. For instance:
```
{{ tpl (toYaml .) $ }}
```

charts/nauth/values.yaml
```diff
@@ -145,3 +145,5 @@ crds:
   install: true
   # -- Indicates if Custom Resource Definitions should be kept when a release is uninstalled.
   keep: true
+
+extraResources: []
```
choufraise, line 150: **suggestion**: Add a Helm unit test for at least one extra resource. Kill two birds with one stone by also verifying the usage of `{{ .Release.Name }}` together with the `tpl` usage in my previous comment. This example test currently fails, but it shouldn't after the `tpl` fix.

```
suite: extra resources
templates:
  - extra_resources.yaml
release:
  name: nauth-test
tests:
  - it: renders extra resources
    set:
      extraResources:
        - apiVersion: v1
          kind: ConfigMap
          metadata:
            name: '{{ .Release.Name }}'
    asserts:
      - equal:
          path: metadata.name
          value: nauth-test
```

charts/nauth/values.yaml
```diff
@@ -145,3 +145,5 @@ crds:
   install: true
   # -- Indicates if Custom Resource Definitions should be kept when a release is uninstalled.
   keep: true
+
```
choufraise, line 148: **suggestion**: Add a `# --` description for this field, and then regenerate `charts/nauth/README.md`.

choufraise (changes requested):
Thanks for the contribution. This is a useful addition for the NAuth chart.

I’m good with the direction, but I’d like the review comments addressed before we merge. Once those are fixed, this should be straightforward to approve.

choufraise:
@madestreel , please also ensure that the commit(s) are signed-off (please see [CONTRIBUTING.md](https://github.com/WirelessCar/nauth/blob/main/CONTRIBUTING.md))

madestreel:
@choufraise Hi, thanks for the suggestions. I did the required changes and signed the commits

choufraise (approved):

choufraise:
@madestreel , sorry for the inconvenience, but may I please ask you to rebase this branch on top of `origin/main`? 🙏  We will need to fix the PR workflow for external PRs, see https://github.com/WirelessCar/nauth/issues/331, but I can't say when we'll find the time to fix this.

Issue #340, opened 2026-07-29
observability: support structured operator logging


choufraise:
## Problem

NAuth operator logs are currently emitted in text-oriented formats that are hard for observability platforms to consume consistently.

Most log lines include timestamp, level, logger, message, and fields as plain text:

~~~text
2026-07-29T07:48:03Z    INFO    Starting workers    {"controller": "accountimport", "controllerGroup": "nauth.io", "controllerKind": "AccountImport", "worker count": 1}
~~~

Some logs use a different klog-style format:

~~~text
I0729 07:47:44.794363       1 leaderelection.go:250] "Attempting to acquire leader lease" lock="nats/949483fb.nauth.io"
~~~

This makes logs harder to process reliably:

- messages include logging prefixes
- levels are inconsistent
- structured fields are embedded in plain text
- source timestamps are embedded in the log body

## Expected outcome

NAuth should support logging configuration suitable for structured log ingestion.

Expected configuration possibilities:

- log format, for example `json` or `console`
- log level, for example `debug`, `info`, `warn`, `error`
- production-friendly defaults for deployed environments

Structured logs should allow observability platforms to consume clean messages, levels, timestamps, and useful fields without an NAuth-specific text parser.

PR #342, opened 2026-07-29
feat: add configurable structured operator logging

choufraise:
Add opt-in JSON logging for NAuth while preserving the existing human-readable text output by default. Expose log format and level configuration through both the operator flags and Helm values, with validation to catch invalid chart settings before deployment.

When JSON logging is enabled, route Kubernetes/client-go klog output through the same logger so leader-election and library messages are emitted consistently. Update the observability guide with concise logging and metrics usage examples.

Refs: #340

### DIFF
--- a/charts/nauth/templates/verify.yaml
+++ b/charts/nauth/templates/verify.yaml
@@ -1,0 +1,6 @@
+{{- if and .Values.logging.format (not (mustHas .Values.logging.format (list "text" "json"))) }}
+{{- fail "logging.format must be empty, text, or json" }}
+{{- end }}
+{{- if and .Values.logging.level (not (mustHas .Values.logging.level (list "debug" "info" "warn" "error"))) }}
+{{- fail "logging.level must be empty, debug, info, warn, or error" }}
+{{- end }}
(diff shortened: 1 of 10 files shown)


charts/nauth/templates/verify.yaml
```diff
@@ -0,0 +1,6 @@
+{{- if and .Values.logging.format (not (mustHas .Values.logging.format (list "text" "json"))) }}
+{{- fail "logging.format must be empty, text, or json" }}
```
cmoscofian, line 2: **nit**: I would not be surprised if someone used the string "empty" based on this message.

charts/nauth/templates/verify.yaml
```diff
@@ -0,0 +1,6 @@
+{{- if and .Values.logging.format (not (mustHas .Values.logging.format (list "text" "json"))) }}
```
cmoscofian, line 1: **suggestion**: starting to feel like **nauth** would benefit from having a values schema.

cmoscofian (approved):

charts/nauth/templates/verify.yaml
choufraise: Empty wouldn't make sense, and the operator wouldn't even start if it was supplied. I'll take that risk tbh. But yeah, you are probably right... 😄

charts/nauth/templates/verify.yaml
choufraise: Issue created https://github.com/WirelessCar/nauth/issues/343

Issue #345, opened 2026-07-30
Investigate unused Helm chart values


choufraise:
NAuth's Helm chart exposes values that are currently documented but not rendered by the chart.

Known values:
- `serviceAccount.automount`
- `podLabels`

We should decide whether each value should be rendered by the chart or removed from the public values API.

This was found while preparing Helm values documentation cleanup ahead of introducing `values.schema.json`.

choufraise:
See [comment regarding this](https://github.com/WirelessCar/nauth/pull/346#pullrequestreview-4818355060) from @cmoscofian 

> though: IMO https://github.com/WirelessCar/nauth/issues/345 we should remove these values and introduce then when(if) required. This can be source of confusion for users, based on existing conventions users will set values expecting certain behaviors. Documenting it is a first step, but since they have no practical use and it's not a safe keyword that needs to be reserved, I would just remove it.

Issue #343, opened 2026-07-30
Add Helm values schema validation for NAuth


choufraise:
NAuth’s Helm chart currently has no `values.schema.json`.

This means invalid values, wrong value types, and simple typos in chart configuration are only caught late, if they are caught at all. A values schema would give users earlier feedback during normal Helm workflows such as linting, templating, installing, and upgrading.

We should investigate and add schema validation for the NAuth chart values while keeping the chart value API easy to maintain.

PR #346, opened 2026-07-30
docs: clarify Helm chart values

choufraise:
Prepare the NAuth Helm values documentation for values schema validation.

Also document currently unused chart values so they can be resolved in a follow-up.

Refs: #343
Refs: #345

### DIFF
--- a/charts/nauth/README.md
+++ b/charts/nauth/README.md
@@ -4,44 +4,41 @@
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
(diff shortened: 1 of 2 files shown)


cmoscofian (approved):
**though**: IMO #345 we should remove these values and introduce then when(if) required. This can be source of confusion for users, based on existing conventions users will set values expecting certain behaviors. Documenting it is a first step, but since they have no practical use and it's not a safe keyword that needs to be reserved, I would just remove it.

PR #347, opened 2026-07-30
docs: update generated CRD reference

choufraise:
Refresh the generated CRD documentation to include AccountSigningKey and signing key reference fields.

### DIFF
--- a/www/src/content/docs/crds.md
+++ b/www/src/content/docs/crds.md
@@ -61,10 +61,10 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
-| `name` _string_ | Name the child resource name |  | MinLength: 1 <br />Required: \{\} <br />Required: \{\} <br /> |
-| `uid` _[UID](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.36/#uid-types-pkg)_ | UID of the child resource UID |  | Required: \{\} <br /> |
-| `observedGeneration` _integer_ | ObservedGeneration refers to the observed generation of the child resource. |  | Minimum: 0 <br />Required: \{\} <br /> |
-| `status` _[AccountAdoptionStatus](#accountadoptionstatus)_ | Status of the adoption |  | Required: \{\} <br /> |
+| `name` _string_ | Name the child resource name |  | MinLength: 1 <br />Required <br /> |
+| `uid` _[UID](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.36/#uid-types-pkg)_ | UID of the child resource UID |  | Required <br /> |
+| `observedGeneration` _integer_ | ObservedGeneration refers to the observed generation of the child resource. |  | Minimum: 0 <br />Required <br /> |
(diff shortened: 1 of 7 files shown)


www/src/content/docs/crds.md
```diff
@@ -456,6 +458,134 @@ _Appears in:_
...
+An Account trusts the public key by listing this resource in Account.spec.signingKeyRefs.
+
+
+
+_Appears in:_
+- [AccountSigningKeyList](#accountsigningkeylist)
+
+| Field | Description | Default | Validation |
+| --- | --- | --- | --- |
+| `apiVersion` _string_ | `nauth.io/v1alpha1` | | |
```
cmoscofian, line 471: **issue**: **apiVersion** and **kind** are duplicated fields in this table. It has the value in the description field in one and another with the description but no value and an optional Validation? This got a bit confusing.

www/src/content/docs/crds.md
```diff
@@ -456,6 +458,134 @@ _Appears in:_
...
+
+AccountSigningKeyList contains a list of AccountSigningKey.
+
+
+
+
+
+| Field | Description | Default | Validation |
+| --- | --- | --- | --- |
+| `apiVersion` _string_ | `nauth.io/v1alpha1` | | |
```
cmoscofian, line 490: **issue**: same as described in the **AccountSigningKey**

www/src/content/docs/crds.md
```diff
@@ -477,6 +607,7 @@ _Appears in:_
 | `imports` _[Imports](#imports)_ |  |  | Optional: \{\} <br /> |
 | `jetStreamLimits` _[JetStreamLimits](#jetstreamlimits)_ |  |  | Optional: \{\} <br /> |
 | `natsLimits` _[NatsLimits](#natslimits)_ |  |  | Optional: \{\} <br /> |
+| `signingKeyRefs` _[AccountSigningKeyRef](#accountsigningkeyref) array_ | SigningKeyRefs lists references whose public keys are trusted as additional<br />signing keys for this account. The implicit default signing key is always<br />present and is not listed here. |  | MaxItems: 64 <br />Optional: \{\} <br /> |
```
cmoscofian: **issue**: This has two validations one with MaxItems and another with Optional: {}

**question**: what does this **Optional: {}** represents?

cmoscofian (commented):
As I was looking at this, I can see that the duplicated fields issue is something that is already present in the other references. Would be nice to adjust everything, since we are fixing docs.

www/src/content/docs/crds.md
choufraise: Optional comes from the `+optional` marker, making it optional of course. This is _not_ conflicting with the `MaxItems`, which is only applicable if the array exists. I'll have a look to see if we can clean up the ugly `Optional: {}`, I agree it doesn't look good.

choufraise: This has been fixed by md template overrides, not pretty in code, but hopefully we will never need to go back to this, and the end result looks much cleaner. I'm 50/50 to this commit. What do you think @cmoscofian ? 

Added as separate commit.

cmoscofian: Looks a lot better 👏🏼

cmoscofian (approved):

PR #348, opened 2026-07-30
feat: add Helm values schema validation

### LINKED ISSUE #343: Add Helm values schema validation for NAuth

choufraise:
Generate values.schema.json from values.yaml annotations and verify it stays in sync through the Helm test workflow.

Closes: #343


.github/workflows/helm-test.yaml
```diff
@@ -27,8 +29,32 @@ jobs:
...
+      - name: Setup mise tools
+        uses: jdx/mise-action@v4
+        with:
+          install: true
+          install_args: "helm"
+          experimental: true
+
+      - name: Verify Helm Values Schema
+        run: |
+          mise run nauth:check-values-schema
```
cmoscofian, line 41: **suggestion**: the plugin maintain their own github action. I would try and take advantage of that.
https://github.com/losisin/helm-values-schema-json#github-actions

Also, it's not super important, but I would keep these separated. I would have one action that handles the schema (generating and validating against the api coming from the latest PR), and another for tests.

Lastly, I'm pretty sure helm runs schema validation by default when you run unittests, templates and linting. So, no need for extra validations on templating as long as tests cover the usecases.

charts/nauth/templates/verify.yaml
cmoscofian, line 1: **praise**: nice housekeeping 🧹

charts/nauth/values.yaml
```diff
@@ -47,42 +54,52 @@ nameOverride: ""
...
   # -- Operator log level. Supported values: `debug`, `info`, `warn`, `error`.
   level: ""
 
+# @schema additionalProperties: false
 serviceAccount:
   # -- Create a ServiceAccount for the NAuth operator.
   create: true
   # TODO(#345): Decide whether to render this value or remove it from the public values API.
   # -- Reserved value for ServiceAccount token automounting. This value is not currently rendered by the chart.
   automount: true
```
cmoscofian, line 80: **suggestion**: one major benefit of this particular plugin against [dadav](https://github.com/dadav/helm-schema) is that it add the possibility to hide fields from the schema.
Per #345 this field is not currently used so would be nice to remove from the schema entirely at this point.

```suggestion
# TODO(#345): Decide whether to render this value or remove it from the public values API.
# @schema hidden
# -- Reserved value for ServiceAccount token automounting. This value is not currently rendered by the chart.
automount: true
```

charts/nauth/values.yaml
```diff
@@ -47,42 +54,52 @@ nameOverride: ""
...
   annotations: {}
   # -- Override the ServiceAccount name. Defaults to the generated full name when empty.
   nameOverride: ""
 
+# @schema type: [object, null]; additionalProperties: {type: string}
 # -- Annotations to add to the NAuth operator Pod.
 podAnnotations: {}
 
 # TODO(#345): Decide whether to render this value or remove it from the public values API.
+# @schema type: [object, null]; additionalProperties: {type: string}
```
cmoscofian, line 92: **suggestion**: same as **serviceAccount.automount**.
```suggestion
# @schema hidden
```

charts/nauth/values.yaml
```diff
@@ -96,13 +113,16 @@ securityContext:
...
+# @schema additionalProperties: false
 monitoring:
   # -- Expose controller-runtime Prometheus metrics on `/metrics` for direct scraping or collection through a Prometheus receiver.
   enabled: false
+  # @schema additionalProperties: false
   serviceMonitor:
     # -- Create Prometheus Operator ServiceMonitor and PrometheusRule resources. Requires the ServiceMonitor and PrometheusRule CRDs.
     enabled: false
 
+# @schema type: [object, null]; skipProperties: true
```
cmoscofian, line 125: **thought**: here we could be extra nice at a small cost and some assumptions we could add a $ref to the kubernetes schema, this would allow users to get instant validation and autocompletion on k8s resources object, and this could be extended to securityContext, probes, volumes and so on.

But there are two possible tradeoffs, we pinpoint a k8s for the api and an external dependency to the docs, but it is common practice.

```suggestion
# @schema  $ref: https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v1.34.0/_definitions.json#/definitions/io.k8s.api.core.v1.ResourceRequirements

```

cmoscofian (dismissed):

.github/workflows/helm-test.yaml
choufraise: The reason I've done what I've done is to ensure that the schema is up to date with the values file. The workflow checks that the generated schema matches the one on the branch. The helm-values-schema-json action always uses the _latest_ version of the plugin, which would break this check. The workflow requires it to be 100% reproducible. Keeping a pinned version in mise ensures this independent on time and machine. 

However, I agree that we should split this into to different jobs and also remove the extra validations. Will do this.

charts/nauth/values.yaml
choufraise: I tested this locally and (sadly) it causes `helm lint` to fail because those defaults still exist in `values.yaml` while the root objects have `additionalProperties: false`. I'll ensure to work on those (#345) as soon as this PR is merged instead.

charts/nauth/values.yaml
choufraise: I'd like to postpone this and keep this PR focused on introducing the schema framework, mainly since it would pin a k8s version to the API.

cmoscofian (approved):

Issue #359, opened 2026-09-10
Account reports Ready when NATS rejects individual imports during claim activation


henriropp:
**Describe the bug**
#### Problem

An `Account` can report `Ready=True` even though NATS rejected one or more imports while activating its JWT.

NATS may accept and store an Account JWT at the request level while logging errors for individual imports, such as:

> `service import not authorized`  
> `stream import not authorized`  
> `import forms a cycle`  
> `account has issues importing account`

NAuth treats a successful claim-update response as complete reconciliation. It records the desired claims in status and marks the `Account` ready without checking whether all imports became effective.

This creates a discrepancy between three states:

- The imports declared in Kubernetes.
- The imports encoded in the stored Account JWT.
- The imports actually active in NATS.

In the observed case, the JWT contained the desired imports, but the effective NATS Account was incomplete and some imports were absent. The Kubernetes resource continued to report success.

#### Impact

Cross-account JetStream sources can stop transferring messages while all related NAuth resources appear healthy. Operators receive no indication from `Account`, `AccountImport`, or `AccountExport` status that the account is incomplete.

This is particularly serious because monitoring based on Kubernetes readiness reports a healthy configuration while the messaging path is unavailable.

#### Expected behavior

`Account` readiness should distinguish between successfully submitting a JWT and NATS successfully activating the complete account configuration.

If NATS rejects or omits any desired import, the Account should not report fully reconciled or ready. Its status should expose enough information to identify the affected import and the reason reported by NATS.

#### Criticality

**High.** This can cause a silent cross-account messaging outage with false-positive health status. There is no confirmed data loss, so it does not currently warrant Critical severity.

choufraise:
One related concern for this issue is the role of `ClaimsHash`. It was introduced to prevent redundant JWT uploads during multi-account deployments, but it is currently treated as an indefinite indication that NATS remains in the expected state.

This connects this ticket with #35 and #352: the JWT can disappear, require an explicit resync, or remain present while its effective account state is incomplete.

I think `ClaimsHash` should remain a write-suppression mechanism, but it should have a bounded validation TTL. Once the TTL expires, NAuth should perform a lightweight check of the current JWT and effective account state, including `Complete`, and only trigger full reconciliation if drift or incompleteness is detected. This preserves the protection against deployment bursts without trusting the hash indefinitely.

For this issue specifically, `Ready=True` should require `Complete=True`; JWT publication/acceptance can be represented separately if useful.

[...]

PR #363, opened 2026-09-14
fix: detect NATS Account JWT drift during reconciliation

### LINKED ISSUE #361: Account reconciliation cannot detect NATS drift when ClaimsHash is unchanged
## Problem

NAuth currently uses `status.claimsHash` to decide whether an Account JWT needs to be uploaded. This prevents redundant uploads when repeated reconciliations produce the same desired claims, but it also causes NAuth to treat an unchanged hash as proof that NATS still has the expected state.

That assumption is invalid. The JWT may be deleted, lost during resolver recovery, or changed outside NAuth while the Kubernetes resource and its `ClaimsHash` remain unchanged. Subsequent reconciliations repeat the same hash comparison and skip the upload, leaving the Account out of sync indefinitely.

The hash is useful for identifying changes in desired claims, but it cannot by itself represent the current state of NATS.

This is related to:

- [#35](https://github.com/WirelessCar/nauth/issues/35), which requires automatic detection and repair of drift between NAuth and NATS.
- [#352](https://github.com/WirelessCar/nauth/issues/352), which proposes explicit resynchronization after known recovery events.
- [#359](https://github.com/WirelessCar/nauth/issues/359), where reliable re-observation of NATS state is a semi-blocker for accurately reporting Account health. The effective-state and readiness semantics remain part of #359.
- [#230](https://github.com/WirelessCar/nauth/pull/230), which introduced the current ClaimsHash-based upload suppression.

## Desired outcome

Account reconciliation must distinguish between:

1. The desired claims calculated by NAuth.
2. Whether those desired claims were recently validated against NATS.
3. Whether the remote NATS state still matches the desired state.

A stable ClaimsHash should continue to prevent redundant uploads while the remote state is considered freshly validated. It must not suppress validation indefinitely. A change in desired claims must still trigger reconciliation immediately, regardless of validation freshness.

## Acceptance criteria

- A matching ClaimsHash is not treated as permanent proof that NATS is up to date.
- A change in desired claims triggers an Account update without waiting for validation expiry.
- Repeated reconciliations with unchanged desired claims do not cause redundant JWT uploads while validation remains fresh.
- Once validation expires, NAuth checks the remote Account state and detects missing or mismatched JWT data.
- Detected drift causes corrective reconciliation.
- Validation freshness is tracked independently from the general reconciliation timestamp.
- Validation failures are surfaced instead of being treated as confirmation that the remote state is unchanged.
- The optimization introduced by #230 remains effective during deployment bursts and repeated child-resource reconciliations.
- The design remains compatible with the explicit resynchronization controls proposed in #352.
- Tests cover unchanged claims, changed claims, expired validation, deleted remote JWTs, remote drift, validation failures, and dependent resources producing the same effective claims.

subhangi-bhardwaj:
## Summary

Improve Account reconciliation so that a matching `ClaimsHash` is not treated as permanent proof that the remote NATS Account is up to date.

This change:

- Tracks the last successful NATS Account state validation using `stateValidatedAt`.
- Reconciles immediately when the desired Account state changes.
- Avoids redundant JWT uploads while the validated state remains fresh.
- Re-checks the remote Account when validation becomes stale.
- Detects and corrects missing or mismatched Account JWT data.
- Preserves the validation timestamp when NATS lookup or upload fails.
- Adds the configurable `account-reconciliation-interval` with a default of five minutes.
- Uses the configured interval for Account validation freshness and periodic reconciliation.
- Preserves the existing optimization for repeated reconciliations and deployment bursts.
- Applies consistent `state` terminology across the CRD, models, controllers, and results.
- Keeps the design compatible with future explicit Account resynchronization controls.
- Leaves the existing Account JWT upload optimization from #230 effective.
- Provides test coverage for unchanged and changed claims, expired validation, deleted or drifted remote JWTs, validation failures, and repeated reconciliations with unchanged effective claims.

Closes: [#361](https://github.com/WirelessCar/nauth/issues/361)

## Testing

- `make test`
- `make lint`
- `go test ./...`
- `make test-e2e`


api/v1alpha1/account_types.go
```diff
@@ -137,6 +137,9 @@ type AccountStatus struct {
 	// ClaimsHash is a hash of the Account JWT claims, used to determine if the claims have changed and a new JWT needs to be generated.
 	// +optional
 	ClaimsHash string `json:"claimsHash,omitempty"`
+	// NatsAccountClaimsValidatedAt records when the desired Account JWT was last validated against NATS.
+	// +optional
+	NatsAccountClaimsValidatedAt metav1.Time `json:"natsAccountClaimsValidatedAt,omitempty"`
```
choufraise: **suggestion**: As you already are in the "Account" (CRD) context, you should avoid over-specifying the field name. It's already implicitly pointing to both "NATS" and "Account", hence simply `ClaimsValidatedAt` would be the preferred name.

api/v1alpha1/account_types.go
```diff
@@ -137,6 +137,9 @@ type AccountStatus struct {
 	// ClaimsHash is a hash of the Account JWT claims, used to determine if the claims have changed and a new JWT needs to be generated.
 	// +optional
 	ClaimsHash string `json:"claimsHash,omitempty"`
+	// NatsAccountClaimsValidatedAt records when the desired Account JWT was last validated against NATS.
```
choufraise: **nit**: "was last validated against NATS" is somewhat ambiguous. It _could_ point to either:
- Validated using the NATS JWT Go library
- Validated against the NATS cluster by being _successfully accepted_ by the NATS resolver
- Validated against the NATS cluster by being _successfully loaded_ by the NATS server(s)

Based on the upcoming Issue #359 we will need to improve the communication between NAuth and NATS, including verifying the `Complete` status. This should be taken into consideration when both naming this field, and describing it in the CRD.

internal/adapter/inbound/controller/account.go
```diff
@@ -205,6 +205,9 @@ func (r *AccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 	natsAccount.Status.Adoptions = adoptions
 	natsAccount.Status.ClaimsHash = result.ClaimsHash
+	if result.ClaimsValidated {
+		natsAccount.Status.NatsAccountClaimsValidatedAt = metav1.Now()
+	}
```
choufraise: **issue**: This is a pure readability issue. When/if the `result.ClaimsValidated` is absent, nil or `False`, we will not update the `natsAccount.Status` object. How should this be interpreted? Does `ClaimsValidated` mean that it was _validated_ (not telling if the validation was successful or not) or should it be interpreted as "validated and accepted by the NATS cluster"? 

Please also note that you are _modifying_ the existing `natsAccount.Status.NatsAccountClaimsValidatedAt`, meaning this timestamp will never be reset, i.e. once set, it will remain that value until overwritten by this if statement. This may or may not be your intention, either way I'd like to point it out.

internal/core/account.go
```diff
@@ -22,6 +22,8 @@ type AccountManager struct {
 	secretManager   secretManager
 }
 
+const accountClaimsValidationInterval = 5 * time.Minute
```
choufraise: **question**: Have you thought about making this hard-coded validation interval a variable at e.g. operator and/or account level with 5 minutes as default? Based on the cluster-size/NAuth usage the users might want to be able to control this as it will affect the NATS cluster and account observability.

internal/core/account.go
```diff
@@ -173,21 +175,17 @@ func (a *AccountManager) CreateOrUpdate(ctx context.Context, request nauth.Accou
...
-		log.Info("Uploaded Account JWT to NATS",
-			"accountID", accountPublicKey, "prevClaimsHash", prevClaimsHash, "claimsHash", claimsHash)
+	claimsValidated, err := a.reconcileAccountJWT(
+		ctx,
+		cluster,
+		accountPublicKey,
+		signedJwt,
+		claimsHash,
+		request.ClaimsHash,
+		request.NatsAccountClaimsValidatedAt,
```
choufraise: **suggestion**: As the`request` (`nauth.AccountRequest`) is a domain model, you could supply the full `request` instead as a parameter so that you can utilize the needed values during reconciliation instead of passing them separately. Then you wouldn't need to ensure name sync between `request.NatsAccountClaimsValidatedAt` and `lastValidation time.Time` in the `reconcileAccountJWT` method signature. It's just a suggestion based on personal preferences.

internal/core/account.go
```diff
@@ -200,9 +198,66 @@ func (a *AccountManager) CreateOrUpdate(ctx context.Context, request nauth.Accou
...
+		}
+
+		if remoteJWT != "" {
+			remoteClaimsHash, err := hashSignedAccountJWTClaims(remoteJWT)
+			if err != nil {
+				return false, fmt.Errorf("failed to validate remote account jwt: %w", err)
+			}
+			if remoteClaimsHash == desiredClaimsHash {
+				log.Info("Validated Account JWT in NATS because claims are unchanged",
+					"accountID", accountID, "claimsHash", desiredClaimsHash)
```
choufraise: **suggestion**: Now when we no longer rely on `ClaimsHash` being changed, we expect NAuth to reconcile the Account much more often. This exact place will become the _preferred end-goal_ of a e.g. all scheduled reconciliations, hence we might end up here a lot. Therefor, it's hard to say if it's worth INFO-logging that "Everything is as it should be!" as that might introduce a lot of unnecessary logs. 

My recommendation is therefor that we should either change this into a debug log, or remove the log entirely due to it being a no-op. If removed, please feel free to add a comment instead like `// Do nothing. Account already in sync with NATS cluster` or similar.

internal/domain/nauth/account.go
```diff
@@ -8,18 +8,19 @@ import (
...
-	JetStreamEnabled *bool                 `json:"jetStreamEnabled,omitempty"`
-	JetStreamLimits  *JetStreamLimits      `json:"jetStreamLimits,omitempty"`
-	NatsLimits       *NatsLimits           `json:"natsLimits,omitempty"`
-	ExportGroups     ExportGroups          `json:"exportGroups,omitempty"`
-	ImportGroups     ImportGroups          `json:"importGroups,omitempty"`
-	SigningKeys      []string              `json:"signingKeys,omitempty"`
+	AccountRef                   domain.NamespacedName `json:"accountRef,omitempty"`
+	AccountID                    AccountID             `json:"accountId,omitempty"`
+	ClaimsHash                   string                `json:"claimsHash,omitempty"`
+	NatsAccountClaimsValidatedAt time.Time             `json:"natsAccountClaimsValidatedAt,omitempty"`
```
choufraise: **issue**: Same as the naming issue I had in the Account CRD, keep the field name focused, e.g. `ClaimsValidatedAt` at tops.

internal/core/account.go
```diff
@@ -173,21 +175,17 @@ func (a *AccountManager) CreateOrUpdate(ctx context.Context, request nauth.Accou
...
-		}
-		defer sysConn.Disconnect()
-
-		err = sysConn.UploadAccountJWT(signedJwt)
-		if err != nil {
-			return nil, fmt.Errorf("failed to upload account jwt: %w", err)
-		}
-		log.Info("Uploaded Account JWT to NATS",
-			"accountID", accountPublicKey, "prevClaimsHash", prevClaimsHash, "claimsHash", claimsHash)
+	claimsValidated, err := a.reconcileAccountJWT(
```
choufraise: **issue**: Same ambiguousness as already mentioned. It's not clear that the `bool` returned from `reconcileAccountJWT` is "Claims validated". It's rather "Claims (successfully) uploaded". 

Note: The claims has already been "validated" long before we get to this point in the manager. In fact, if line 164 (`natsClaims, err := claimsBuilder.build()`) is not erroneous it means that the claims has been validated successfully by the (local) NATS JWT library itself.

choufraise (commented):
Looks promising. However, I have some comments mainly regarding to naming that makes it a bit ambiguous to read the code between e.g. CRD, controller and manager.

internal/core/account.go
subhangi-bhardwaj: Made the validation interval configurable at the operator level through `nats.accountClaimsValidationInterval / --account-claims-validation-interval,` with 5m as the default. This will allow operators  to tune the NATS lookup frequency based on cluster size and usage, account-level overrides are not currently included.

charts/nauth/templates/deployment.yaml
```diff
@@ -36,6 +36,7 @@ spec:
             - --health-probe-bind-address=:8081
             - --metrics-bind-address=:8080
             - --metrics-secure=false
+            - --account-claims-validation-interval={{ .Values.nats.accountClaimsValidationInterval }}
```
choufraise, line 39: Thanks for iterating on this. I also want to correct my earlier feedback: I focused on `ValidatedAt` as if the timestamp wording was the problem. That was the wrong part to focus on, and I’m sorry for sending you in that direction.

Your original implementation was already very close to the right concept. The important distinction is that this represents the overall Account state, not only the Account claims. `stateValidatedAt` captures that much better.

**suggestion**: Let’s rename this to `account-reconciliation-interval` everywhere, including the Helm value and internal variables.

The setting should control the behavior its name promises. Pass it to the `Account` controller and use it when scheduling the next periodic reconciliation. With the current hard-coded [`RequeueAfter` of approximately five minutes](https://github.com/WirelessCar/nauth/blob/91ecfa36806830034dcc5ee8b0ff2698ee630b1a/internal/adapter/inbound/controller/account.go#L218), configuring `account-reconciliation-interval` to `1m` would not have the expected effect.

The same value should also be used as the freshness interval for NATS validation:

- On an event-driven reconciliation, skip the NATS call when the desired Account state has not changed and `status.stateValidatedAt` is still within the configured interval.
- Contact NATS when the timestamp is stale or when the desired Account state, currently represented by `ClaimsHash`, has changed.
- Schedule the next periodic Account reconciliation using the configured interval.

For this PR, `status.stateValidatedAt` should mean “the time when NAuth last successfully validated the desired Account state in NATS”. Update it only after a successful Account JWT upload or after a lookup confirms that the remote JWT matches the desired state. If NATS returns an error, leave the timestamp unchanged so that the state remains stale and will be retried.

Please apply the same `state` terminology to the related CRD, model, result, and comment names. A lookup confirms the remote Account state; it does not mean that claims were newly accepted, so names such as `ClaimsAcceptedAt` and `ClaimsAcceptanceConfirmed` are misleading.

The `stateValidatedAt` name also remains suitable for the future `Complete` check, because it refers to the Account’s overall state rather than only its claims. Implementing the `Complete` check is not required for this PR, but if you believe covering both changes together would be suitable and make the implementation clearer, please feel free to do so.

internal/adapter/inbound/controller/account.go
choufraise: This will be solved when we've improved the field name. Resolving.

api/v1alpha1/account_types.go
choufraise: See [super seeding comment](https://github.com/WirelessCar/nauth/pull/363#discussion_r4006269726) (my focus was a bit off here). Closing.

internal/core/account.go
```diff
@@ -173,36 +180,84 @@ func (a *AccountManager) CreateOrUpdate(ctx context.Context, request nauth.Accou
...
+			return false, fmt.Errorf("failed to validate account jwt in NATS: %w", err)
+		}
+
+		if remoteJWT != "" {
+			remoteClaimsHash, err := hashSignedAccountJWTClaims(remoteJWT)
+			if err != nil {
+				return false, fmt.Errorf("failed to validate remote account jwt: %w", err)
+			}
+			if remoteClaimsHash == desiredClaimsHash {
+				// The remote JWT already matches the desired claims.
```
choufraise, line 244: **suggestion**: Since we do have a Verbose log at line 221 (the other no-op path), we should probably be consistent and have a verbose log here as well.

charts/nauth/tests/deployment_account_reconciliation_interval_test.yaml
```diff
@@ -0,0 +1,18 @@
+suite: account reconciliation interval args
+templates:
+  - deployment.yaml
+tests:
+  - it: uses the five minute reconciliation interval by default
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --accountReconciliationInterval=5m
```
choufraise: **suggestion**: This parameter should be kebab-cased, but the adjacent parameter in values should still be camelCased.

internal/adapter/inbound/controller/constants.go
```diff
@@ -67,3 +67,7 @@ const ( // "requeue after" durations
 	// Poll interval when a required dependency exists but is not yet ready.
 	requeueDependencyNotReady = 5 * time.Second
 )
+
+// DefaultAccountReconciliationInterval is the default period used for Account
+// reconciliation and NATS Account state validation freshness.
+const DefaultAccountReconciliationInterval = 5 * time.Minute
```
choufraise: **suggestion**: Since this is the only default value that is made public, to be used by `main.go`, we should most probably move this constant out of `internal` and put it into `cmd/main.go` since that's the only place where we set the final Account Reconciliation Interval value.

choufraise (commented):
Starting to look really great! Some minor comments left.

internal/adapter/inbound/controller/status.go
```diff
@@ -49,11 +49,19 @@ func (s *statusReporter) status(ctx context.Context, object Object) (ctrl.Result
 	}
 
 	// Spreading out the requeue to avoid all being queued at the same time
+	// TODO: Make the status requeue interval configurable instead of hard-coding five minutes.
```
choufraise: **praise**: Nice that we got a TODO on this old hard-coded stuff. 👍

choufraise (approved):
Nicely done! Please ensure to update the PR summary (and commit message if needed) according to the final implementation before merging.

PR #366, opened 2026-09-17
docs(adr): document NAuth interpolation syntax context

choufraise:
## Summary

Add ADR-6 to document the need for a reserved, NAuth-wide interpolation syntax.

The ADR covers:

- Reconciliation-time and Auth Callout rendering values.
- Helm compatibility and fail-closed behavior.
- NATS subject validation and NATS-owned template syntax.
- Candidate interpolation patterns and their trade-offs.
- Additional established syntaxes considered.


choufraise (commented):
@henriropp had a suggestion to a different approach to this due to the possible risks mentioned in the ADR. 

Based on the idea from NATS JWT design of the `TokenPosition` for subject `mappings`, we could utilize the CRD structure to make this strict without defining an NAuth "global" substitution pattern.

The following example is based on `AccountImport` CRD but should be applicable to `AccountExport`, `User` (`permissions`) and other future CRD's requiring subject substitutions, like `Role` etc. 
```
AccountImport:
  spec:
    rules:
      # Alternative 1 - Named token substitution
      - subjectTpl:
          template: $JS.FC._.{{ ACCOUNT_HASH }}.my-stream.>
          substitutions:
            - token: {{ ACCOUNT_HASH }}
              value: ExportAccountHash
      # Alternative 2 - Indexed substitutions
      - subjectTpl:
          template: $JS.FC._.*.my-stream.>
          substitutions:
            - position: 3
              value: ExportAccountHash
```
The `ExportAccountHash` would be an NAuth internal, well-known enum-like string per context and/or CRD.

One shared consequence, in comparison to the substitution syntax described in this ADR, is that this is quite more verbose and might make it harder to get an overview of e.g. a larger `ImportGroup`, `ExportGroup` or `User` spec.

### Alternative 1:
- The preferred token name/pattern does not need to be set, users may use whatever token string they'd like, e.g. `sub.{{ AccountID }}.>`, `sub.AID.>`, `sub.<%AccountID>.>`; matching`token` requires a perfect string match.
Cons:
- Supports straight forward multi-token substitutions, e.g. `sub.{{AccountID}}.str.{{AccountID}}.>` 
- Supports trivial re-arranging of subject tokens in the template without the need to update any index references. 

### Alternative 2:
- Follows the existing NATS `PositionToken` indexing pattern.
- In a template like `*.*.*.*.>` it might be hard to to know which token `position: 2` will replace based on if it's zero- or one based positioning. This is error prone.

adr/0006-nauth-interpolation-syntax.md
```diff
@@ -0,0 +1,154 @@
...
+- Conventional spacing, as in `{{ .nauth.account.hash }}`, makes an unresolved marker contain ASCII spaces, so NATS rejects it.
+- Familiar to Go and Helm users.
+- Supports dotted field paths rooted in `.nauth`.
+- Existing Go template tooling can support implementation.
+
+**Cons**
+
+- Requires explicit Helm escaping and has two rendering stages.
+- Go also accepts whitespace-free forms such as `{{.nauth.account.hash}}`; those remain valid NATS subjects, so NAuth must validate unresolved markers independently and cannot rely on whitespace universally.
+- Shares delimiters with NATS-owned templates. NAuth must preserve only recognized NATS expressions in fields where NATS supports them, treat only `{{ .nauth.<field> }}` as NAuth syntax, and reject all other `{{...}}` content.
```
kejne, line 90: **issue:** This is quite a major flaw and likely breaks the entire idea of using go templating. This would require you to essentially hand-roll your own logic to identify the go template syntax and only template this.

kejne (commented):
**praise:** Nicely written ADR. I think you covered the important aspects and identified a big issue with the otherwise given go templating approach 👍

adr/0006-nauth-interpolation-syntax.md
choufraise: Yes, that is a big draw back with this option.

choufraise:
Separate but related update: I’m also investigating whether this functionality could/should be implemented in NATS rather than NAuth. The proposal is for NATS to support interpolation in Account JWT import/export subjects, for example:
```
$JS.FC._.{{account-hash()}}.stream-a.>
```

NATS would resolve the hash from the referenced account for imports and the current account for exports during Account loading. If viable and accepted upstream, this would remove the need for NAuth-side substitution for the current FC V2 use case. I’m keeping this ADR open while I validate that possibility with the NATS/Synadia maintainers.

PR #370, opened 2026-09-21
fix(account): require observed NATS completeness for readiness

### LINKED ISSUE #359: Account reports Ready when NATS rejects individual imports during claim activation

choufraise:
## Summary

An Account could previously become `Ready` after its JWT was uploaded, even when NATS had not loaded the Account or resolved its imports.

This PR makes readiness reflect the observed NATS Account state.

## Changes

- Observe Account completeness through `ACCOUNTZ`.
- Add the `NATSAccountComplete` condition.
- Keep `Ready` false when NATS reports an incomplete Account.
- Keep both `NATSAccountComplete` and `Ready` as `Unknown` when completeness cannot be established.
- Request runtime Account loading before completeness validation.
- Refresh incomplete Account JWTs before retrying validation.
- Persist the observed NATS state and validation timestamp in Account status.
- Add unit, regression, and E2E coverage using a three-node NATS cluster.
- Document the readiness behavior and observation workflow.
- Add [ADR-6](https://github.com/WirelessCar/nauth/blob/main/adr/0006-account-readiness-single-server-observation.md), defining single-server NATS observation.

The implementation intentionally observes one connected NATS server per validation. It does not query every server in a cluster.

## Commits

- `e087da8` — `docs(adr): define single-server Account readiness observation`
- `3035115` — `test(account): preserve existing readiness behavior`
- `4be0287` — `feat(nats): observe Account completeness through ACCOUNTZ`
- `e38c04f` — `fix(account): gate readiness on observed NATS completeness`
- `9e434a9` — `docs(www): document Account readiness validation workflow`
- `397a9fe` — `fix(account): materialize Accounts before NATS completeness validation`
- `8a45faf` — `test(e2e): verify asynchronous Account readiness convergence`

Fixes #359


api/v1alpha1/account_types.go
```diff
@@ -138,11 +153,9 @@ type AccountStatus struct {
 	// Account state has changed and a new JWT needs to be generated.
 	// +optional
 	ClaimsHash string `json:"claimsHash,omitempty"`
-	// StateValidatedAt records when NAuth last successfully validated the desired Account state
-	// in NATS, either through a successful Account JWT upload or a matching Account lookup. It
-	// is not updated when validation is skipped or fails.
+	// Nats contains NATS-specific Account state observation evidence.
 	// +optional
-	StateValidatedAt metav1.Time `json:"stateValidatedAt,omitempty"`
+	Nats *AccountNatsStatus `json:"nats,omitempty"`
```
choufraise, line 158: **note**: Since `StateValidatedAt` has not yet been released, this is not considered a breaking API change.

test/e2e/account-import-conflict/01-create-accounts.yaml
```diff
@@ -15,3 +15,6 @@ spec:
   natsClusterRef:
     namespace: nats
     name: local-nats
+  exports:
```
choufraise, line 18: **note**: Export is needed for the account to become Complete in NATS, hence adding this to ensure the test case is still valid.

local/nats/values.yaml
```diff
@@ -15,7 +16,8 @@ nats:
       enabled: true
       merge:
         type: full
-        interval: 2m
+        # Keep local resolver convergence below the KUTTL step timeout.
+        interval: 5s
```
choufraise, line 20: **note**: Each NATS server periodically checks another server for Account JWT updates and shares any differences. The 5s interval helps this test cluster pick up updates within KUTTL's short deadlines. Such a short interval is only needed for testing; production clusters should use a longer interval to avoid unnecessary background traffic.

test/e2e/account-deletion-jetstream/01-assert.yaml
```diff
@@ -4,6 +4,9 @@ metadata:
   name: example-account
 status:
   conditions:
+    - type: NATSAccountComplete
```
choufraise, line 7: **note**: Since KUTTL verifies the full `conditions` list, we need to add this new condition to all existing tests as well.

choufraise (commented):
Left some clarification notes for reviewers.

adr/0006-account-readiness-single-server-observation.md
```diff
@@ -0,0 +1,52 @@
...
+Query the server reached by the NATS connection with an identity-matched ACCOUNTZ request and treat the result as representative of the cluster.
+
+### Option 2 - Observe multiple or all servers
+
+Collect several ACCOUNTZ responses and define a confidence threshold.
+This requires server discovery, threshold configuration, larger responses, and additional reconciliation cost.
+
+## Decision
+
+For the first implementation level, use Option 1.
```
henriropp: This signals that there is a second.
> For the first implementation level,

**Suggestion**: Remove that phrase.

internal/adapter/inbound/controller/account.go
```diff
@@ -312,6 +343,93 @@ func toBootstrapAccountRequest(state *v1alpha1.Account, accountReference nauth.A
...
+	invalidImportCount := 0
+	for _, imp := range state.Imports {
+		if !imp.Invalid {
+			continue
+		}
+		invalidImportCount++
+		if len(invalidImports) == maxInvalidImports {
+			continue
+		}
+		invalidImports = append(invalidImports, fmt.Sprintf("%s -> %s (%s)", imp.AccountID, imp.Subject, imp.Type))
```
henriropp: **suggestion**: I first read this to only skip the 5th import from being appended. But I now see it will work, but maybe use `invalidImportCount >= maxInvalidImports` ?

or reverse it:
```
if len(invalidImports) < maxInvalidImports {
  invalidImports = append(invalidImports, fmt.Sprintf("%s -> %s (%s)", imp.AccountID, imp.Subject, imp.Type))
}
```

internal/adapter/inbound/controller/account.go
```diff
@@ -312,6 +343,93 @@ func toBootstrapAccountRequest(state *v1alpha1.Account, accountReference nauth.A
...
+	}
+	if len(invalidImports) == 0 {
+		return "NATS reports the Account as incomplete"
+	}
```
henriropp: check: is this correct?

internal/adapter/outbound/nats/account_load.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"encoding/json"
+	"fmt"
+)
+
+const (
+	accountLoadRequestSubject = "$SYS.REQ.ACCOUNT.NSUBS"
+	// accountLoadSubject is an arbitrary literal required by ACCOUNT.NSUBS while calculating
+	// subscription counts; NAuth ignores the count and uses the Account lookup as the side effect.
+	accountLoadSubject = "x"
```
henriropp, line 12: You did not go with what we discussed 😉 ? (internal joke)

henriropp (commented):
Good work! Just some minor comments.

internal/adapter/inbound/controller/account.go
choufraise: Yes, it would be the default, generic Incomplete message. But I will restructure this a bit and put this default message at the end of the function instead. It's weird to first look into imports, then defaults, and then go back to "blaming" imports once again. Great finding!

internal/adapter/outbound/nats/account_load.go
choufraise: Haha no, I did not. 🤣  `hey.baberiba` or `easter.egg` felt _too_ childish/unprofessional.

internal/adapter/inbound/controller/account.go
choufraise: Fixed in [updated commit](https://github.com/WirelessCar/nauth/pull/370/changes/5e3b39a3c7f64e2ee64e4f4ef4d2d0f5645ace98). Resolving.

internal/adapter/inbound/controller/account.go
choufraise: Fixed in [updated commit](https://github.com/WirelessCar/nauth/pull/370/changes/5e3b39a3c7f64e2ee64e4f4ef4d2d0f5645ace98). @henriropp , please have a second look.

adr/0006-account-readiness-single-server-observation.md
choufraise: Fixed in [updated commit](https://github.com/WirelessCar/nauth/pull/370/commits/5e877fb3653b9e1b8858006a0fb369b6d011c18d). @henriropp , please have a second look.

henriropp (approved):
Nice!

PR #372, opened 2026-09-23
fix(account): report imports missing from NATS observations

choufraise:
## Summary

When NATS reports an Account as incomplete, imports that could not be materialized may be absent from its observed import list. NAuth previously showed only a generic incomplete message, leaving users without a useful lead when an expected export was missing.

Compare desired Account JWT imports with observed imports and name desired imports that were not reported. The status describes only what was observed; it does not infer why an import is absent. NATS-reported invalid imports continue to be shown.

Adds regression coverage for NATS reporting an invalidated import and a readiness assertion for a desired import omitted from the observation.

Related to #359

## Commits
- **test(nats): verify ACCOUNTZ reports invalidated imports**
- **fix(account): report imports missing from NATS observations**


henriropp (approved):
LGTM

